PR #1, opened 2018-10-31
[WIP] Cloudtrail parser

ajvb:
Initial attempt at a Cloudtrail parser.

In its current form this is of limited usefulness as it does not support Stackdriver, but wanted to create this PR and get some initial review before I continue on. Also have two TODO comments that I'll call out as I'm not sure the best way to resolve them.

Also, the whole way in which I am doing JSON parsing for Cloudtrail is very much in question. After trying to use a few different third-party packages, none of them were allowing for the simplicity or the number of fields this way supports.


src/main/java/com/mozilla/secops/parser/Cloudtrail.java
```diff
@@ -0,0 +1,96 @@
...
+    public Cloudtrail(String input, Event e, Parser p) {
+        try {
+            event = mapper.readValue(input, CloudtrailEvent.class);
+            if (isAuthEvent()) {
+              Normalized n = e.getNormalized();
+              n.setType(Normalized.Type.AUTH);
+              n.setSubjectUser(getUser());
+              n.setSourceAddress(getSourceAddress());
+              //TODO
+              //n.setObject();
```
ajvb: Just saw the updated comment in `Normalized` explaining what this is. Will update.

src/main/java/com/mozilla/secops/parser/GLB.java
```diff
@@ -41,6 +41,10 @@ public Boolean matcher(String input) {
             // pass
         } catch (IllegalArgumentException exc) {
             // pass
+        } catch (NullPointerException exc) {
```
ajvb: @ameihm0912 As the comment below states, when the raw Cloudtrail logs were the `input` in this function, it was getting to `String eType = (String)m.get("@type");` and throwing a NullPointerException.

Maybe the proper fix is to just not support raw cloudtrail events and just support them within Stackdriver?

ameihm0912: The null pointer exception should be fixed in f2a7e50

src/main/java/com/mozilla/secops/parser/Cloudtrail.java
```diff
@@ -0,0 +1,97 @@
...
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+
+import com.mozilla.secops.parser.models.cloudtrail.CloudtrailEvent;
+
+public class Cloudtrail extends PayloadBase implements Serializable {
+    private static final long serialVersionUID = 1L;
+
+    private final ObjectMapper mapper = getObjectMapper();
```
ameihm0912: I'm not sure if we would want this initialized globally like this. The object needs to be serializable to be passed around in the beam pipeline. It would probably be better to allocate this as needed in whatever local function requires it (`matcher` or the event constructor) so we aren't passing it around in each event.

src/main/java/com/mozilla/secops/parser/Cloudtrail.java
```diff
@@ -0,0 +1,97 @@
...
+
+    private CloudtrailEvent event;
+
+    @Override
+    public Boolean matcher(String input) {
+        try {
+          event = mapper.readValue(input, CloudtrailEvent.class);
+          if (Double.parseDouble(event.getEventVersion()) > 1.00) {
```
ameihm0912: Are there any other fields we could use to be more accurate here?

This is kind of a drawback in the current parser as the higher level (e.g., stackdriver/encapsulation) probably indicates better what type of event it is but we don't have access to it here. This is probably fine for now and we can improve it later.

src/main/java/com/mozilla/secops/parser/Cloudtrail.java
```diff
@@ -0,0 +1,97 @@
...
+
+import java.io.IOException;
+import java.io.Serializable;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+
+import com.mozilla.secops.parser.models.cloudtrail.CloudtrailEvent;
+
+public class Cloudtrail extends PayloadBase implements Serializable {
```
ameihm0912: Could we add basic javadoc comments around all of the public classes and functions?

src/main/java/com/mozilla/secops/parser/Cloudtrail.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    private ObjectMapper getObjectMapper() {
+        ObjectMapper _mapper = new ObjectMapper();
+        _mapper.disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+        return _mapper;
+    }
+
+    public String getSourceAddress() {
+      return event.getSourceIPAddress();
```
ameihm0912: I think a bunch of this indentation is out of wack in here

src/main/java/com/mozilla/secops/parser/GLB.java
```diff
@@ -41,6 +41,10 @@ public Boolean matcher(String input) {
...
+            // TODO: Cloudtrail test gets to `(String)m.get("@type")`
+            //       and causes a null pointer exception.
+            // pass
```
ameihm0912: As mentioned we can probably take this extra catch out once f2a7e50 is merged

src/main/java/com/mozilla/secops/parser/models/cloudtrail/CloudtrailEvent.java
```diff
@@ -0,0 +1,127 @@
...
+  private String eventVersion;
+  private Boolean readOnly;
+  private String recipientAccountID;
+  private String requestID;
+  private String sourceIPAddress;
+  private String userAgent;
+
+  public UserIdentity userIdentity;
+
+  public HashMap<String, Object> responseElements;
```
ameihm0912: Does this stuff and `userIdentity` need to be public for some reason?

src/main/java/com/mozilla/secops/parser/models/cloudtrail/CloudtrailEvent.java
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  public String getIdentityName() {
+    if (getUserType().equals("IAMUser")) {
+      return userIdentity.userName;
+    } else if (getUserType().equals("AssumedRole")) {
+      try {
+        String username = userIdentity.sessionContext.sessionIssuer.get("userName");
+        return username;
+      } catch (NullPointerException exc) {
```
ameihm0912: It's better here to check the required values for null, we should avoid catching NPE here since we can validate the values before hand.

src/main/java/com/mozilla/secops/parser/models/cloudtrail/CloudtrailEvent.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    return null;
+  }
+
+  public Boolean wasMFAAuthenticated() {
+    try {
+      String mfad = userIdentity.sessionContext.attributes.get("mfaAuthenticated");
+      return Boolean.parseBoolean(mfad);
+    } catch (NullPointerException e) {
```
ameihm0912: Avoid catching `NullPointerException`

src/main/java/com/mozilla/secops/parser/models/cloudtrail/CloudtrailEvent.java
```diff
@@ -0,0 +1,127 @@
...
+      return null;
+    }
+  }
+
+  public Object getResponseElements(String key) {
+    try {
+      Object o = responseElements.get(key);
+      return o;
+    } catch (NullPointerException e) {
```
ameihm0912: Avoid catching `NullPointerException`

src/main/java/com/mozilla/secops/parser/models/cloudtrail/SessionContext.java
```diff
@@ -0,0 +1,8 @@
+package com.mozilla.secops.parser.models.cloudtrail;
+
+import java.util.HashMap;
+
+public class SessionContext {
```
ameihm0912: Does this need to be public or is this class just used internally within the object?

It might be better to have all member access happen through the Cloudtrail class and avoid making all these values public in the package.

src/main/java/com/mozilla/secops/parser/models/cloudtrail/UserIdentity.java
```diff
@@ -0,0 +1,12 @@
+package com.mozilla.secops.parser.models.cloudtrail;
+
+public class UserIdentity {
```
ameihm0912: Similar to previous comment here, private with private members and getter's I think would be cleaner?

src/main/java/com/mozilla/secops/parser/models/cloudtrail/CloudtrailEvent.java
ajvb: Oh yeah, I meant to come back to this. Was having trouble with Jackson not parsing out into these Objects unless they were `public` so just left them `public` at the time.

ajvb:
@ameihm0912 Believe this is ready for re-review. Added support for Stackdriver unwrapping + used identitymanager for normalize.

src/main/java/com/mozilla/secops/parser/Cloudtrail.java
```diff
@@ -0,0 +1,201 @@
...
+     *
+     * @return Source address country string
+     */
+    public String getSourceAddressCountry() {
+        return sourceAddressCountry;
+    }
+
+
+    private Boolean isAuthEvent() {
+        if (event.getEventName().equals("ConsoleLogin")) {
```
ameihm0912: Do we need null checks here if the event didn't have an event name?

ajvb: Yeah, added.

PR #16, opened 2018-11-28
Cloudtrail Pipeline

ajvb:
Port of `moz_security_cloudtrail`.

There are a few things left to do to have feature parity (regex field matchers, improve alert message), but wanted to get a PR up to get some review sooner rather than later.


src/main/java/com/mozilla/secops/awsbehavior/AwsBehavior.java
```diff
@@ -0,0 +1,172 @@
...
+    private static final long serialVersionUID = 1L;
+
+    @Override
+    public PCollection<Event> expand(PCollection<String> col) {
+      EventFilter filter = new EventFilter();
+      filter.addRule(new EventFilterRule().wantSubtype(Payload.PayloadType.CLOUDTRAIL));
+
+      return col.apply(ParDo.of(new ParserDoFn()))
+          .apply(EventFilter.getTransform(filter))
+          .apply(Window.<Event>into(FixedWindows.of(Duration.standardMinutes(5))));
```
ameihm0912: Do we want fixed windows here? I believe with this windowing strategy, the data in a given window will only be emitted once the watermark has past the end of the window, so it could delay events by up to 5 minutes.

It might be better to leave the events in the global window and use a trigger to ensure the panes are fired more rapidly (e.g., every 10 seconds or something).

src/main/java/com/mozilla/secops/awsbehavior/AwsBehavior.java
```diff
@@ -0,0 +1,172 @@
...
+
+    private Logger log;
+
+    public Matcher(CloudtrailMatcher cm) {
+      this.cm = cm;
+      log = LoggerFactory.getLogger(Matcher.class);
+      this.filter = new EventFilter();
+      try {
+        this.filter.addRule(cm.toEventFilterRule());
+      } catch (CloudtrailMatcher.UnknownStringPropertyException exc) {
```
ameihm0912: With this catch it will basically result in the `Matcher` being established with an `EventFilter` with no rules in it

This probably wont cause problems as I think that should just drop every event but I am wondering if we should add something into the expand function to just abort if something went wrong in the constructor?

src/main/java/com/mozilla/secops/awsbehavior/AwsBehavior.java
```diff
@@ -0,0 +1,172 @@
...
+                    }
+                  }));
+    }
+  }
+
+  public static class Matchers extends PTransform<PCollection<Event>, PCollection<Alert>> {
+    private static final long serialVersionUID = 1L;
+
+    private final String cmmanagerPath;
+    private CloudtrailMatcherManager cmmanager;
```
ameihm0912: I'm not 100% sure but I think we might want to make sure `CloudtrailMatcherManager` is serializable if it's being setup in the constructor

src/main/java/com/mozilla/secops/awsbehavior/AwsBehavior.java
```diff
@@ -0,0 +1,172 @@
...
+      }
+      PCollection<Alert> ret = alerts.apply(Flatten.<Alert>pCollections());
+      return ret;
+    }
+  }
+
+  /** Runtime options for {@link AwsBehavior} pipeline. */
+  public interface AwsBehaviorOptions extends PipelineOptions, InputOptions, OutputOptions {
+    @Description("Identity manager configuration; resource path")
+    @Default.String("/identitymanager.json")
```
ameihm0912: Could we put this into a directory? e.g., `/awsbehavior/identitymanager.json`

src/main/java/com/mozilla/secops/awsbehavior/AwsBehavior.java
```diff
@@ -0,0 +1,172 @@
...
+    String getIdentityManagerPath();
+
+    void setIdentityManagerPath(String value);
+
+    @Description("Cloudtrail matcher manager configuration; resource path")
+    @Default.String("/event_matchers.json")
```
ameihm0912: Likewise here

src/main/java/com/mozilla/secops/parser/Cloudtrail.java
```diff
@@ -197,4 +196,41 @@ private Boolean isAuthEvent() {
...
+    switch (resource) {
+      case "requestParameters.userName":
+        HashMap<String, Object> rp = event.getRequestParameters();
+        Object u = rp.get("userName");
+        if (u == null) {
+          return "";
+        }
+        return (String) u;
+    }
+    return "";
```
ameihm0912: Should this return a null or something if it fails instead of `""`?

src/main/java/com/mozilla/secops/awsbehavior/AwsBehavior.java
ameihm0912: Actually, scratch this -- this is the utility class right and not part of the pipeline itself? This is probably OK here then since we would want this to be shared

src/main/java/com/mozilla/secops/awsbehavior/AwsBehavior.java
ameihm0912: https://github.com/mozilla-services/foxsec-pipeline/blob/master/src/main/java/com/mozilla/secops/authprofile/AuthProfile.java#L73-L80

src/main/java/com/mozilla/secops/awsbehavior/AwsBehavior.java
ajvb: Yeah, maybe this should be moved into a constant within IdentityManager?

ameihm0912: https://github.com/mozilla-services/foxsec-pipeline/pull/18

PR #20, opened 2018-12-05
Add support for regex matchers in EventFilterPayload

ajvb:


src/main/java/com/mozilla/secops/parser/EventFilterPayload.java
```diff
@@ -84,6 +97,18 @@ public Boolean matches(Event e) {
...
 
+  /**
+   * Add a new string regex match to the payload filter
+   *
+   * @param property {@link EventFilterPayload.StringProperty}
+   * @param s String regex pattern to match against
+   * @return EventFilterPayload for chaining
+   */
+  public EventFilterPayload withStringRegexMatch(StringProperty property, String s) {
+    stringRegexMatchers.put(property, Pattern.compile(s));
```
ameihm0912: What happens here if the regex is invalid?

PR #23, opened 2018-12-13
Email HTML Templates

ajvb:
HTML templates for emails with an initial one for AuthProfile

screenshot_20181213_113257

(The fact that the later text is light grey is due to google groups threading / repeated messages)


src/main/java/com/mozilla/secops/alert/TemplateManager.java
```diff
@@ -0,0 +1,38 @@
...
+
+  /**
+   * Create html email body using supplied template name and template variables
+   *
+   * @param templateName Name of template file
+   * @param variables Variables for template
+   * @return Assembled email body
+   */
+  public String createEmailBody(String templateName, HashMap<String, Object> variables)
+      throws Exception {
```
ameihm0912: Is it possible to be more specific about the type(s) of exceptions we are expecting here?

src/main/java/com/mozilla/secops/authprofile/AuthProfile.java
```diff
@@ -175,6 +179,11 @@ public void processElement(ProcessContext c) throws StateException {
         }
 
         Alert alert = new Alert();
+
+        // TODO: Should this be moved?
+        // TODO: Should the template only be used at a certain severity?
+        alert.setTemplateName("authprofile.ftlh");
```
ameihm0912: The output messages have different content depending on if it is a new location or not; I am guessing we will need two templates here.

I think hardcoding the template name is fine here, maybe we could do something like if setTemplateName() can't find the required template only the text body is output.

src/main/java/com/mozilla/secops/authprofile/AuthProfile.java
```diff
@@ -224,6 +233,32 @@ public void processElement(ProcessContext c) throws StateException {
         if (country != null) {
           alert.addMetadata("sourceaddress_country", country);
         }
+        if (city != null && country != null) {
+          alert.addMetadata("geolocation", String.format("%s, %s", city, country));
```
ameihm0912: This is already in the metadata in `sourceaddress_{city,country}` so guessing this is here just for the template? Should we just always add these metadata fields and default them to "unknown" so the template can always find them or something if not set rather then duplicate information in the metadata?

src/main/java/com/mozilla/secops/authprofile/AuthProfile.java
```diff
@@ -224,6 +233,32 @@ public void processElement(ProcessContext c) throws StateException {
...
+            CityResponse cr = geoip.lookup(ip);
+            if (cr != null) {
+              location =
+                  String.format("%s, %s", cr.getCity().getName(), cr.getCountry().getIsoCode());
+            }
+            currentEntries.add(
+                String.format(
+                    "%s - %s - %s", entry.getValue().getTimestamp(), entry.getKey(), location));
+          }
+          alert.addMetadata("current_entries", String.join("\n", currentEntries));
```
ameihm0912: I'd like to avoid putting linefeeds into metadata since it is meant to be accessed directly and not really parsed

For now could we maybe change this to something like previous_sourceaddress_country, previous_sourceaddress_city, etc and just store it like that? We could just take the latest known entry from state.

src/main/java/com/mozilla/secops/alert/Alert.java
```diff
@@ -266,4 +287,18 @@ public String toJSON() {
...
   }
+
+  /**
+   * Return HashMap used by Freemarker to generate an HTML alert email
+   *
+   * @return Template data model
+   */
+  public HashMap<String, Object> generateTemplateVariables() {
+    HashMap<String, Object> v = new HashMap<String, Object>();
+    v.put("alert", this);
```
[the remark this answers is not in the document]
ajvb: Yes, but you can call getters from within the templates using a `.field_name` syntax. In the template it calls `alert.timestamp` and `alert.alert_id` using this.

src/main/java/com/mozilla/secops/authprofile/AuthProfile.java
ajvb: >The output messages have different content depending on if it is a new location or not; I am guessing we will need two templates here.

Oh true. I think for now we should just support HTML email for a new location.

>I think hardcoding the template name is fine here, maybe we could do something like if setTemplateName() can't find the required template only the text body is output.

We do this within `AlertMailer.createAlertMailBody` - https://github.com/mozilla-services/foxsec-pipeline/pull/23/files#diff-971a7614812a70496be0ef5724e089ddR117

src/main/java/com/mozilla/secops/authprofile/AuthProfile.java
ajvb: Make perfect sense. Let me refactor this out of here / not store it in the alert metadata.

ajvb:
@ameihm0912 This should be ready for a re-review.

I decided to just rip out the `current_entries` section for now, as it would require a bit more refactoring than I think is worth it at the moment since we are not even sure if we are going to use this.

PR #42, opened 2019-01-17
Slack support for alert notifications

ajvb:
* Adds functionality for sending alerts to a Slack "catch all" channel
* Adds most functionality for sending alerts with confirmation buttons to users in Slack.


src/main/java/com/mozilla/secops/alert/AlertIO.java
```diff
@@ -100,6 +105,22 @@ public void processElement(ProcessContext c) {
...
+        if (cfg.getSlackCatchall() != null) {
+          // Configured catchall slack channel always recieves a copy of the alert
+          if (!slack.sendToCatchall(a)) {
+            log.error("failed to send alert to slack catchall");
+          }
+        }
+
+        String slackEmail = a.getMetadataValue("notify_slack_direct");
+        if (slackEmail != null) {
+          if (!slack.confirmationAlert(a, slack.getUserId(slackEmail))) {
```
ameihm0912: Will these every be purely informational, in the sense `confirmationAlert` wouldn't apply?

Should we have a different metadata key if its a confirmation vs. just an informational direct message?

src/main/java/com/mozilla/secops/alert/AlertSlack.java
```diff
@@ -0,0 +1,108 @@
...
+  public String getUserId(String email) {
+    if (emailToSlackUserId == null) {
+      // TODO: Move this to IdentityManager
+      try {
+        emailToSlackUserId = slackManager.getEmailToUserIdMapping();
+      } catch (IOException exc) {
+        log.error("error getting user list from slack (IOException): {}", exc.getMessage());
+      } catch (SlackApiException exc) {
+        log.error("error getting user list from slack (SlackApiException): {}", exc.getMessage());
+      }
```
ameihm0912: Should we return here if this fails and make sure `emailToSlackUserId` is still null so we can try again next time? It looks like if this fails the first time it will not recover and `emailToSlackUserId` will just stay an empty map.

src/main/java/com/mozilla/secops/authprofile/AuthProfile.java
```diff
@@ -217,6 +217,10 @@ public void processElement(ProcessContext c) throws StateException {
               log.info(
                   "{}: adding direct email notification metadata route to {}", identityKey, dnot);
               alert.addMetadata("notify_email_direct", dnot);
+
+              // TODO: Uncomment to enable slack direct message alerts
```
ameihm0912: Can we just enable this now? It will only actually take effect if the output transform is configured for slack alerting right?

src/main/java/com/mozilla/secops/OutputOptions.java
```diff
@@ -55,6 +55,16 @@
 
   void setOutputAlertEmailCatchall(String value);
 
+  @Description("With alert slack output; Slack token (supports RuntimeSecrets)")
```
ameihm0912: @ajvb one other comment here, it doesn't look like this crypto class is actually implemented for use with this parameter?

src/main/java/com/mozilla/secops/alert/AlertIO.java
ajvb: I'd say that at the moment they will never be purely informational. Easy to change later if this changes, but I do not see us wanting to send alerts to users unless we are asking them to interact with them.

src/main/java/com/mozilla/secops/authprofile/AuthProfile.java
ajvb: We can and yes, but it also means we can't use the slack catchall channel on real AuthProfile until this is completed. I'm impartial, but at the same time it might be nice to keep in uncommented just in case so we don't send broken alerts to users on accident. (Broken because the buttons won't actually do anything right now)

src/main/java/com/mozilla/secops/alert/AlertIO.java
ameihm0912: 👌

src/main/java/com/mozilla/secops/authprofile/AuthProfile.java
ameihm0912: @ajvb ok makes sense; could we just remove the commented code/TODO from the PR for now then before we put it in master? We can just add the line needed once we are ready?

ajvb: :+1: absolutely

PR #55, opened 2019-01-28
Improvements to AuthProfile

ameihm0912:
A bunch of improvements to AuthProfile and supporting classes:

* Add new AUTH_SESSION normalized event type
* Modify GcpAudit parser to avoid storing non-serializable AuditLog
* Add state delete all function, primarily for use in tests to clear existing state
* Improve AuthProfile tests
* In AuthProfile, GBK using identity string to avoid state update collisions


ameihm0912:
> Is the use of `AUTH_SESSION` for GCP because we do not get direct auth events?

Yeah, idea is rather than look for a specific authentication event we extract information from the existing session on audited API calls

PR #64, opened 2019-01-31
Use normalized fields in HTTPRequest

ameihm0912:
* Add support for processing a second variant of nginx SD jsonPayload
* Support filtering events in `EventFilter` using Stackdriver labels
* Normalize nginx/GLB payloads to `HTTP_REQUEST`
* Converts HTTPRequest pipeline to operate on `HTTP_REQUEST` normalized fields


ameihm0912:
@ajvb yeah right now we cannot consume both in the same pipeline at the same time

PR #72, opened 2019-02-07
cidr fetch from gcs, support processing multi-address nginx source fields

ameihm0912:


src/main/java/com/mozilla/secops/parser/Parser.java
```diff
@@ -45,6 +46,46 @@ public static DateTime parseISO8601(String in) {
...
+  /**
+   * Apply any configured multi-address selector to the specified input string
+   *
+   * <p>If an HTTP multi-address selector was not configured in the parser configuration, the input
+   * is just returned as is.
+   *
+   * @param input Input string
+   * @return Results of address selector application
+   */
+  public String applyHttpMultiAddressSelector(String input) throws IllegalArgumentException {
```
ajvb: Is this needed only when `ngx_http_realip_module` isn't configured correctly?

ajvb: If so, should the configuration for this be more like [ngx_http_realip_module](https://nginx.org/en/docs/http/ngx_http_realip_module.html) where trusted IPs/CIDRs are set?

ameihm0912:
@ajvb this should be good for a second look now

PR #75, opened 2019-02-14
Alert state, direct slack alert support

ajvb:
Adds
* Support for alert persistence in either memcached or datastore
* Support for sending alerts directly to users

Note: Moving the memcached and datastore options will require changes to the deployment options


ajvb:
@ameihm0912 should be ready for another look, separated the state opts.

src/main/java/com/mozilla/secops/OutputOptions.java
```diff
@@ -71,6 +72,29 @@
...
+  void setAlertStateMemcachedPort(Integer value);
+
+  @Description("Use datastore alert state; namespace for entities")
+  @Default.String("alerts")
+  String getAlertStateDatastoreNamespace();
+
+  void setAlertStateDatastoreNamespace(String value);
+
+  @Description("Use datastore alert state; kind for entities")
+  @Default.String("alerts")
```
ameihm0912: If we are going to remove the boolean for datastore should we also remove the defaults then so it's not enabled by default for datastore?

ajvb: :man_facepalming: yes, fixing

PR #77, opened 2019-02-22
Whitelisted IPs support

ajvb:
~Blocked By: https://github.com/mozilla-services/foxsec-pipeline-contrib/pull/3~
Requires a fix in the [slackbot db](https://github.com/mozilla-services/foxsec-pipeline-contrib/blob/master/common/db.go) code first, as the pipeline's `State` object expects the object to be stored as json within the `state` field.

---

Integrates with the ip whitelist slackbot by marking alerts as "iprepd exempt". Alerts are stored with metadata but not sent to iprepd.


src/main/java/com/mozilla/secops/OutputOptions.java
```diff
@@ -93,6 +93,12 @@
 
   void setAlertStateDatastoreKind(String value);
 
+  @Description("Enable use of whitelisted ips saved in datastore; requires deployment in GCP")
+  @Default.Boolean(false)
+  Boolean getEnableWhitelistedIps();
```
ameihm0912: Since this is specific to iprepd, why don't we call this something like `getOutputIprepdEnableDatastoreWhitelist` so it is consistent with the other iprepd output options?

src/main/java/com/mozilla/secops/state/DatastoreStateInterface.java
```diff
@@ -83,7 +83,10 @@ public void initialize() throws StateException {
     } else {
       datastore = DatastoreOptions.getDefaultInstance().getService();
     }
-    keyFactory = datastore.newKeyFactory().setNamespace(namespace).setKind(kind);
```
ameihm0912: Don't we always want a namespace, otherwise we can end up with unrelated objects in the default namespace? What was the intent around this one?

src/main/java/com/mozilla/secops/whitelistedips/WhitelistedIpsLabeler.java
```diff
@@ -0,0 +1,57 @@
...
+import com.mozilla.secops.alert.Alert;
+import com.mozilla.secops.state.DatastoreStateInterface;
+import com.mozilla.secops.state.State;
+import com.mozilla.secops.state.StateException;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** */
+public class WhitelistedIpsLabeler extends DoFn<String, String> {
```
ameihm0912: I'm not sure on this part of it - if we need a new package for this since it seems like it's primarily related to iprepd.

Rather then adding a whole new `DoFn`, what if we just had a static function in `IprepdIO`, maybe something like `addMetadataIfWhitelisted` or similar.

In the existing `HTTPRequest DoFn` elements, when the alert is being built we could just call this function to add the extra metadata entry if needed. This also has the added benefit of keeping all the metadata processing related to this in one class, since the check would happen there too.

For example:

```java
Alert a = new Alert();
a.addMetadata("sourceaddress", remoteAddress);
if (enableIprepdDatastoreWhitelist) {
    IprepdIO.addMetadataIfWhitelisted(remoteAddress, a);
}
```

This way functions that want to make use of the whitelist do not need to be aware of the metadata tag being used for it.

The JSON model class for the whitelist can just be moved to a private class in the same file as the `IprepdIO` class.

I'm not against the extra `DoFn` completely if we need it however.

src/main/java/com/mozilla/secops/state/DatastoreStateInterface.java
ajvb: This was my bad. I was reading various datastore docs and misunderstood the case when you don't want to use a namespace. Changing the go/slackbot code to require a namespace + reverting this.

src/main/java/com/mozilla/secops/OutputOptions.java
ajvb: :+1:

src/main/java/com/mozilla/secops/whitelistedips/WhitelistedIpsLabeler.java
ajvb: That makes sense to me! Fixing.

PR #78, opened 2019-02-25
Customs conversion for FxaAuth

ameihm0912:
Conversion of existing customs pipeline to operate on FxaAuth data


src/main/java/com/mozilla/secops/parser/EventFilter.java
```diff
@@ -117,6 +119,10 @@ public Boolean matches(Event e) {
   /**
    * Given any keySelectors return the applicable key from the event
    *
+   * <p>Base64 encoding is applied to key sub elements that are used in the returned key.
```
ajvb: Why the base64 encoding?

src/main/resources/customs/customsdefault.json
```diff
@@ -2,7 +2,7 @@
     "detectors": {
```
ajvb: NB: I do wonder if we should make this file private/move it out of the repo.

src/main/java/com/mozilla/secops/parser/EventFilter.java
ameihm0912: To keep it simple we are just using a string here for the key, and the base64 encoding is used to make sure the input elements do not contain whichever delimiter we are using to separate the elements in it. For example, if a space is used to separate keys we don't want spaces in the elements as well as that will cause problems later splitting them.

src/main/resources/customs/customsdefault.json
ameihm0912: Possibly at some point, right now this is just the same data you see in https://github.com/mozilla/fxa-customs-server/blob/master/README.md in the policies section. At some point we will likely want to organize this a bit differently.

PR #84, opened 2019-03-07
Use sentence style in authprofile alert summary

ajvb:
Remove the difference between summary and payload construction for
alerts in authprofile and just use the full sentence style.

Alerts in Slack catchall will now look like:
>An authentication event for user ajvb was detected to access emit-bastion from 127.0.0.1 [unknown/unknown]. This occurred from a source address unknown to the system.  (64979792-cb45-4b9a-b596-48ef4813d2a7)

Alerts to users in Slack will now look like:
>Foxsec Fraud Detection Alert
>
>An authentication event for user ajvb was detected to access emit-bastion from 127.0.0.1 [unknown/unknown]. This occurred from a source address unknown to the system.
>
>If this was not you, or you have any questions about this alert, email us at secops@mozilla.com with the alert id: 648086bd-4564-4fca-9e50-b56f4f9a288c

### DIFF
--- a/src/main/java/com/mozilla/secops/alert/AlertSlack.java
+++ b/src/main/java/com/mozilla/secops/alert/AlertSlack.java
@@ -97,8 +97,8 @@ public class AlertSlack {
 
     String text =
         String.format(
-            "Foxsec Fraud Detection Alert\n\n%s\n%s\nAlert Id: %s",
-            a.getSummary(), a.assemblePayload(), a.getAlertId());
+            "Foxsec Fraud Detection Alert\n\n%s\n\nIf this was not you, or you have any questions about this alert, email us at secops@mozilla.com with the alert id: %s\n",
+            a.getSummary(), a.getAlertId());
     try {
       return slackManager.handleSlackResponse(slackManager.sendMessageToChannel(userId, text));
     } catch (IOException exc) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/mozilla/secops/authprofile/AuthProfile.java
```diff
@@ -378,8 +360,9 @@ public void processElement(ProcessContext c) throws StateException {
           }
         }
 
-        buildAlertSummary(e, a);
-        buildAlertPayload(a);
+        String summary = buildAlertSummary(e, a);
+        a.setSummary(summary);
+        a.addToPayload(summary);
```
ameihm0912: If we are setting the payload to the same thing as the summary, should we just leave the summary field as is here and just send the payload data in direct user notifications for Slack? This would keep the summary concise like the rest of the pipelines.

src/main/java/com/mozilla/secops/alert/AlertSlack.java
```diff
@@ -97,8 +97,8 @@ public Boolean sendToUser(Alert a, String userId) {
 
     String text =
         String.format(
-            "Foxsec Fraud Detection Alert\n\n%s\n%s\nAlert Id: %s",
-            a.getSummary(), a.assemblePayload(), a.getAlertId());
+            "Foxsec Fraud Detection Alert\n\n%s\n\nIf this was not you, or you have any questions about this alert, email us at secops@mozilla.com with the alert id: %s\n",
```
ameihm0912: See other comment

If we just send the payload here, we could append this extra text in the payload in the authprofile pipeline itself.

That way we can support other types of direct notifications here that this extra comment may not apply to.

ajvb: :+1: that sounds good to me. Will open a new PR with that change.

PR #88, opened 2019-03-11
Set IdentityManager for Parser within ParserDoFn

ajvb:
Currently, the mapping from account id -> account name for `Normalized.object` (https://github.com/mozilla-services/foxsec-pipeline/blob/master/src/main/java/com/mozilla/secops/parser/Cloudtrail.java#L75-L86) is not getting set within AuthProfile.

This adds setting the IdentityManager within ParserDoFn with support for passing in the path.

### DIFF
--- a/src/main/java/com/mozilla/secops/parser/ParserDoFn.java
+++ b/src/main/java/com/mozilla/secops/parser/ParserDoFn.java
@@ -57,8 +57,8 @@ public class ParserDoFn extends DoFn<String, Event> {
     }
     log = LoggerFactory.getLogger(ParserDoFn.class);
 
-    IdentityManager mgr = null;
     try {
+      IdentityManager mgr;
       if (idmanagerPath == null) {
         mgr = IdentityManager.load();
       } else {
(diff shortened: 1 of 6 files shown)


src/main/java/com/mozilla/secops/parser/ParserDoFn.java
```diff
@@ -14,6 +16,7 @@
 
   private EventFilter inlineFilter;
   private ParserCfg cfg;
+  private String idmanagerPath;
```
ameihm0912: Can we move this into `ParserCfg`?

e.g., https://github.com/mozilla-services/foxsec-pipeline/blob/0739e2cfd956cfc9c2405c40733d577101ef035c/src/main/java/com/mozilla/secops/parser/ParserCfg.java#L12-L15

Then we can just set it in https://github.com/mozilla-services/foxsec-pipeline/blob/0739e2cfd956cfc9c2405c40733d577101ef035c/src/main/java/com/mozilla/secops/parser/ParserCfg.java#L22 and make it an input option.

This will also require doing the additional setup here:

https://github.com/mozilla-services/foxsec-pipeline/blob/0739e2cfd956cfc9c2405c40733d577101ef035c/src/main/java/com/mozilla/secops/parser/Parser.java#L291

ajvb: :+1:

PR #92, opened 2019-03-18
Add some additional authprofile handling

ameihm0912:


ameihm0912:
> Should we exclude events that match from generating alerts? Or from escalating alerts to users?

I was thinking maybe the latter initially or perhaps we just adjust the alert text to indicate its likely cloud console?

ajvb:
> > Should we exclude events that match from generating alerts? Or from escalating alerts to users?
> 
> I was thinking maybe the latter initially or perhaps we just adjust the alert text to indicate its likely cloud console?

Yeah, I like the idea of just not escalating it to user for now.

PR #96, opened 2019-03-20
httprequest hard limit analyzer

ameihm0912:
Hard fixed window limit analysis with configurable per-source rate


ameihm0912:
> looks great.
> 
> From an user/operator perspective, should HardLimitAnalysis and ThresholdAnalysis be mutually exclusive in that you don't want both running at the same time as they are for different traffic patterns?

yeah, can't see any reason we would run both -- however they are both default disabled, so you'd have to explicitly enable both to get both. maybe there is some sort of use case here?

PR #103, opened 2019-03-28
If using WSL, avoid symlink for volume mounts

ajvb:


bin/m
```diff
@@ -13,6 +13,12 @@ if [[ `uname` == 'Darwin' ]]; then
 	CACHEFLAGS=':cached'
 fi
 
+# If using WSL, the devevloper is likely using a symlink to a directory
```
ameihm0912: small typo here

PR #105, opened 2019-03-28
httprequest ua blacklist

ameihm0912:


src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
```diff
@@ -381,6 +384,136 @@ public void processElement(ProcessContext c, BoundedWindow w) {
...
+                      }
+                      for (Pattern p : uaRegex) {
+                        if (p.matcher(ua).matches()) {
+                          c.output(KV.of(n.getSourceAddress(), ua));
+                          return;
+                        }
+                      }
+                    }
+                  }))
+          .apply(GroupByKey.<String, String>create())
```
ajvb: Should we start naming all transforms?

ameihm0912: @ajvb yeah I think so, maybe we should do a pass over httprequest in another PR

Was thinking we should also do a pass for exceptions that aren't being propagated up properly, I noticed a few cases like that

ajvb: :+1: Both of those sound good

PR #108, opened 2019-03-29
named subnets within idmanager and authprofile

ajvb:
Add support for named subnets within IdentityManager and make use of
those as the entry key within AuthProfile.


src/main/java/com/mozilla/secops/CidrUtil.java
```diff
@@ -57,6 +57,11 @@ public void processElement(ProcessContext c) {
     };
   }
 
+  public static boolean AddressInCidr(String ipAddr, String cidr) {
```
ameihm0912: Could we add some doc comments on this? Also should we use `addressInCidr`?

PR #109, opened 2019-04-01
GCS template loading + alert templates for slack

ajvb:


src/main/java/com/mozilla/secops/GcsUtil.java
```diff
@@ -1,18 +1,47 @@
...
+    }
+    Storage storage = StorageOptions.getDefaultInstance().getService();
+    // check if blob exists
+    Blob blob = storage.get(bid);
+    if (blob == null || !blob.exists()) {
+      return null;
+    }
+    try {
+      return storage.signUrl(BlobInfo.newBuilder(bid).build(), 10, TimeUnit.MINUTES);
+    } catch (IllegalStateException exc) {
```
ameihm0912: Should these be propagated up?

src/main/java/com/mozilla/secops/alert/Alert.java
```diff
@@ -263,22 +263,41 @@ public String getCategory() {
...
   /**
-   * Get template name
+   * Get email template name
    *
    * @return Freemarker template name with file extension or null if not set.
    */
   @JsonIgnore
-  public String getTemplateName() {
-    return getMetadataValue("template_name");
+  public String getEmailTemplateName() {
```
ameihm0912: Will one always be applicable? Do we need some sort of postfix for the type of alert, e.g., direct, etc?

src/main/java/com/mozilla/secops/authprofile/AuthProfile.java
```diff
@@ -251,6 +251,12 @@ private Alert createBaseAlert(Event e) {
         a.addMetadata("sourceaddress_country", "unknown");
       }
 
+      if (e.getNormalized().isOfType(Normalized.Type.AUTH)) {
+        a.addMetadata("auth_alert_type", "auth");
+      } else if (e.getNormalized().isOfType(Normalized.Type.AUTH_SESSION)) {
+        a.addMetadata("auth_alert_type", "auth");
```
ameihm0912: should this be auth_session?

ajvb: yes, thank you!

src/main/java/com/mozilla/secops/alert/Alert.java
ajvb: imo we can add that later if we need it. I imagine that we will only use templates for direct alerts and one template per notification system per pipeline for now.

src/main/java/com/mozilla/secops/GcsUtil.java
ajvb: Hmm, yeah probably. Would be worth logging if auth is failing.

ajvb:
I believe something is up with travis ci. The error is related to a change that is definitely in this branch (`.getTemplateName()` being changed in `TestAuthProfile.java`).

PR #110, opened 2019-04-02
Improve performance of httprequest

ameihm0912:
* Calculate mean in `Stats` `CombineFn`
* Prefer `CombineFn` over `CombineFnWithContext`
* Window output into unified global windows for alerting output
* Remove excessive bundle logging from threshold analysis `DoFn`
* Improve performance of various JSON deserialization functions
* Share parser `ObjectMapper` and `JsonFactory` with payload parsers

### DIFF
--- a/src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
+++ b/src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
@@ -792,17 +792,6 @@ public class HTTPRequest implements Serializable {
                   new DoFn<KV<String, Long>, Alert>() {
                     private static final long serialVersionUID = 1L;
 
-                    private Boolean warningLogged;
-                    private Boolean clampMaximumLogged;
-                    private Boolean statsLogged;
-
-                    @StartBundle
-                    public void startBundle() {
-                      warningLogged = false;
-                      clampMaximumLogged = false;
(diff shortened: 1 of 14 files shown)


src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
```diff
@@ -801,63 +792,23 @@ public void processElement(ProcessContext c) {
...
                     @ProcessElement
                     public void processElement(ProcessContext c, BoundedWindow w) {
                       Stats.StatsOutput sOutput = c.sideInput(wStats);
-                      Long uc = c.sideInput(uniqueClients);
+                      Long uc = sOutput.getTotalElements();
                       Map<String, Boolean> nv = c.sideInput(natView);
 
                       Double cMean = sOutput.getMean();
-                      if (!statsLogged) {
-                        log.info(
```
ajvb: Could these log messages still by useful in a "debug" or "verbose" mode that is not enabled by default?

ameihm0912: Yeah I think they are useful but the way we were doing it (via bundle start) probably isn't a great way to handle this as it can result in 1000's of log messages per second being emitted

It would be good to figure out a proper way to log this information once per window.

PR #118, opened 2019-04-09
transactions for persistent state

ameihm0912:
Improves persistent state handling by utilizing transactions where
possible.

For Datastore based state implementations, transaction support has been
added that can provide error reporting on key contention.

The existing memcached state implementation has been modified to work
with the newer cursor API, however it has been noted that no transaction
support exists in the implementation.

### DIFF
--- a/src/main/java/com/mozilla/secops/state/DatastoreStateCursor.java
+++ b/src/main/java/com/mozilla/secops/state/DatastoreStateCursor.java
@@ -1,0 +1,62 @@
+package com.mozilla.secops.state;
+
+import com.google.cloud.datastore.Datastore;
+import com.google.cloud.datastore.DatastoreException;
+import com.google.cloud.datastore.Entity;
+import com.google.cloud.datastore.Key;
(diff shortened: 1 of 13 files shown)


src/main/java/com/mozilla/secops/state/DatastoreStateCursor.java
```diff
@@ -0,0 +1,63 @@
...
+    Key nk;
+    try {
+      nk = keyFactory.newKey(s);
+    } catch (IllegalArgumentException exc) {
+      throw new StateException(exc.getMessage());
+    }
+    Entity.Builder eb = Entity.newBuilder(nk);
+    eb.set("state", v);
+    Entity e = eb.build();
+    Boolean tryAdd = false;
```
ajvb: looks like `tryAdd` is no longer used?

ameihm0912: Indeed, fixed

PR #121, opened 2019-04-11
Support slack confirmation alerts; Include event timestamp in alert metadata

ajvb:
* Adds support for sending alerts using a confirmation button. (integration/slack code pr: https://github.com/mozilla-services/foxsec-pipeline-contrib/pull/12)
* Adds eventTimestamp() to Normalized and makes use of it in Cloudtrail and OpenSSH

### DIFF
--- a/src/main/java/com/mozilla/secops/alert/AlertSlack.java
+++ b/src/main/java/com/mozilla/secops/alert/AlertSlack.java
@@ -104,6 +104,7 @@ public class AlertSlack {
 
     log.info("generating slack message for {}", userId);
 
+    a.addMetadata("slack_type", "notification");
     String text = createAlertBody(a);
     try {
       return slackManager.handleSlackResponse(slackManager.sendMessageToChannel(userId, text));
@@ -122,15 +123,21 @@ public class AlertSlack {
    * @param userId Slack user id
    * @return Boolean on whether the alert was sent successfully
    */
-  public Boolean confirmationAlert(Alert a, String userId) {
+  public Boolean sendConfirmationAlertToUser(Alert a, String userId) {
     if (a == null || userId == null) {
       return false;
     }
 
     log.info("storing state of alert for {}", userId);
 
+    if (state == null) {
+      log.error("alert state has not been configured");
+      return false;
+    }
+
+    a.addMetadata("slack_type", "confirmation");
+    a.addMetadata("status", "NEW");
     try {
-      a.addMetadata("status", "NEW");
       state.initialize();
       state.set(a.getAlertId().toString(), a);
     } catch (StateException exc) {
@@ -140,9 +147,7 @@ public class AlertSlack {
 
     log.info("generating slack message for {}", userId);
 
-    String text =
-        String.format(
-            "Foxsec Fraud Detection Alert\n\n%s\n\nalert id: %s", a.getPayload(), a.getAlertId());
+    String text = createAlertBody(a);
     try {
       return slackManager.handleSlackResponse(
           slackManager.sendConfirmationRequestToUser(userId, a.getAlertId().toString(), text));
--- a/src/main/java/com/mozilla/secops/alert/AlertSlack.java
+++ b/src/main/java/com/mozilla/secops/alert/AlertSlack.java
@@ -105,7 +105,6 @@ public class AlertSlack {
 
     log.info("generating slack message for {}", userId);
 
-    a.addMetadata("slack_type", "notification");
     String text = createAlertBody(a);
     try {
       return slackManager.handleSlackResponse(slackManager.sendMessageToChannel(userId, text));
@@ -136,8 +135,6 @@ public class AlertSlack {
       return false;
     }
 
-    a.addMetadata("slack_type", "confirmation");
-    a.addMetadata("status", "NEW");
     StateCursor c = null;
     try {
       state.initialize();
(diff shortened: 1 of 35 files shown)


ajvb:
@ameihm0912 It looks like setting the timestamp of `Event` within `Parser.stripStackdriverEncapsulation` (https://github.com/mozilla-services/foxsec-pipeline/pull/121/files#diff-43cb63cf80a3ad715b052542cc3560e3R143) breaks some Customs tests, and results in less than expected matches. i.e.:

```
[ERROR] Failures: 
[ERROR]   TestCustoms.rlMultiTest:160 Combine.globally(Count)/Values/Values/Map/ParMultiDo(Anonymous).output: 
Expected: iterable over [<6L>] in any order
     but: Not matched: <5L>
```

I spent sometime reading through the code, but am a little bit confused as to why this is happening. Is it that the timestamps in https://github.com/mozilla-services/foxsec-pipeline/blob/master/src/test/resources/testdata/customs_multi1.txt are now being used instead of just `time.now()` for all of them?

src/main/java/com/mozilla/secops/alert/AlertSlack.java
```diff
@@ -105,6 +105,7 @@ public Boolean sendToUser(Alert a, String userId) {
 
     log.info("generating slack message for {}", userId);
 
+    a.addMetadata("slack_type", "notification");
```
ameihm0912: I'm wondering, should we add this additional metadata field back in `AlertIO` `processElement` rather than here? The reason being just clarity and reducing how deep in the alerting code we will still modify the alert metadata?

Also, should this potentially be more generic, e.g., something like `alert_notification_type` we just set in `AlertIO` so the templates are usable for notifications besides just Slack output?

src/main/java/com/mozilla/secops/identity/Notify.java
```diff
@@ -28,6 +29,16 @@ public Boolean getDirectSlackNotify() {
     return directSlackNotify;
   }
 
+  /**
+   * Return if identity should be directly notified via Slack using confirmation alert
+   *
+   * @return Boolean to indicate direct notification via Slack using confirmation alert
+   */
+  @JsonProperty("direct_slack_confirmation_notify")
```
ameihm0912: I'm not sure if we need a new field here, can we not just reuse the existing direct notification preferences?

src/main/java/com/mozilla/secops/identity/Identity.java
```diff
@@ -86,16 +86,39 @@ public String getEmailNotifyDirect(Notify defaultNotification) {
...
+    if ((defaultNotification == null)
+        || (defaultNotification.getDirectSlackNotify() != null
+            && defaultNotification.getDirectSlackNotify() == false)) {
+      return false;
+    }
+
+    return true;
+  }
+
+  /**
```
ameihm0912: It seems like we shouldn't need this extra function; the identity configuration would indicate if a direct notification should be generated, but the actual logic of if it's a confirmation should probably live in authprofile?

src/main/java/com/mozilla/secops/parser/Normalized.java
```diff
@@ -19,6 +20,7 @@
 
   private EnumSet<Type> types;
 
+  private DateTime eventTimestamp;
```
ameihm0912: Why not just use the timestamp in `Event` if we need it instead of adding it here as well?

src/main/java/com/mozilla/secops/parser/OpenSSH.java
```diff
@@ -54,16 +60,20 @@ public OpenSSH(String input, Event e, ParserState state) {
...
+      authMethod = mat.group(3);
+      user = mat.group(4);
+      sourceAddress = mat.group(5);
       Normalized n = e.getNormalized();
       n.addType(Normalized.Type.AUTH);
       n.setSubjectUser(user);
       n.setSourceAddress(sourceAddress);
       n.setObject(hostname);
 
+      DateTime et = DateTime.parse(authTimestamp, DateTimeFormat.forPattern("MMM dd HH:mm:ss"));
```
ameihm0912: I think we should prefer setting this in the event instead of normalized, but it would be good to add a new static function to `Parser` similar to `parseISO8601`

src/main/java/com/mozilla/secops/parser/Parser.java
```diff
@@ -140,6 +140,7 @@ private String stripStackdriverEncapsulation(Event e, String input, ParserState
 
       e.setStackdriverProject(getStackdriverProject(entry));
       e.setStackdriverLabels(entry.getLabels());
+      e.setTimestamp(parseISO8601(entry.getTimestamp()));
```
ameihm0912: I think in certain cases the embedded timestamp (e.g., mozlog, textPayload, etc) could be different than what is set by Stackdriver; but this happens earlier than the parser step which can override the event timestamp if need be so is probably fine

src/main/java/com/mozilla/secops/identity/Notify.java
ajvb: Yes, I'd like to do that in the future, but at the moment this is useful for enabling confirmation alerts for a subset of user. Want to turn it on for just the two of us, then SecOps, then everyone.

ajvb:
> @ajvb I think we should split the changes to the parser and events related to timestamps out of this PR and just use the event timestamp (initially) in the alerts; then PR support to extract timestamps out in separate branches; there is quite a bit going on in here that affects everything
> 
> Lets just use `Event.getTimestamp()` anywhere we need a timestamp in the metadata for now and then as we improve the event handling we will get accuracy there?

:+1: That sounds good.

PR #122, opened 2019-04-11
refactor authprofile

ameihm0912:
Refactor some authprofile functionality to make it easier to expand on
the existing analysis path.


ameihm0912:
@ajvb no major functionality changes here, but breaking this out to make review of the follow up PR a bit easier.

PR #124, opened 2019-04-15
Set Event timestamp as time event occurred where possible

ajvb:
Sets Event.timestamp as time event occurred where possible. It defaults
"timestamp" within Stackdriver before setting it based on type-specific
fields.

For OpenSSH, year is not included, so the timestamp of when the event
was received is used to infer year. This could yield sadness on or
around new years midnight.

There are some changes to test data in this PR, as the test data had invalid timestamps.

I ran:
```
sed -i "s/1979-01-01T00:00:00.0000000000/1970-01-01T00:00:00.000000000Z/g" *
```
in `src/test/resources/testdata/` (`s/1979/1970/g` and swapping the last `0` for `Z`).


src/main/java/com/mozilla/secops/parser/Parser.java
```diff
@@ -140,6 +140,14 @@ private String stripStackdriverEncapsulation(Event e, String input, ParserState
 
       e.setStackdriverProject(getStackdriverProject(entry));
       e.setStackdriverLabels(entry.getLabels());
+      if (entry.getTimestamp() != null) {
+        try {
+          e.setTimestamp(Parser.parseISO8601(entry.getTimestamp()));
+        } catch (IllegalArgumentException exc) {
+          // This will occur if the timestamp is not properly formatted. This should only happen in
+          // testing.
+        }
```
ajvb: @ameihm0912 I added this as I thought it would be good for testing live pipelines, but am unsure whether we want to blow up here or not.

src/main/java/com/mozilla/secops/authprofile/AuthProfile.java
```diff
@@ -299,6 +300,11 @@ private Alert createBaseAlert(Event e) {
         a.addMetadata("auth_alert_type", "auth_session");
       }
 
+      DateTime eventTimestamp = e.getTimestamp();
+      if (eventTimestamp != null) {
```
ameihm0912: This should never happen, so we can probably shorten this

```java
a.addMetadata("event_timestamp", e.getTimestamp().toString();
```

Are we OK with the default output time format from `toString`?

Will we want this field in all alerts? If so, it might be nice to just add a new function to the `Alert` class to populate the right metadata with the correct timestamp format, e.g.,:

```java
a.eventTimestampMetadataFromEvent(e);
```

If we want it everywhere too, consider adding a check in https://github.com/mozilla-services/foxsec-pipeline/blob/9f4c4a25fe4a8c94643dfa22066a86e301f0e477/src/main/java/com/mozilla/secops/alert/Alert.java#L57.

src/main/java/com/mozilla/secops/parser/Cloudtrail.java
```diff
@@ -60,6 +60,9 @@ public Cloudtrail(String input, Event e, ParserState state) {
     mapper = getObjectMapper();
     try {
       event = parseInput(input);
+      if (event.getEventTime() != null) {
+        e.setTimestamp(Parser.parseISO8601(event.getEventTime()));
```
ameihm0912: This will overwrite the event timestamp with `null` if for some reason the timestamp can't be parsed.

We should check the return value before applying it here.

src/main/java/com/mozilla/secops/parser/OpenSSH.java
```diff
@@ -54,16 +59,27 @@ public OpenSSH(String input, Event e, ParserState state) {
...
       Normalized n = e.getNormalized();
       n.addType(Normalized.Type.AUTH);
       n.setSubjectUser(user);
       n.setSourceAddress(sourceAddress);
       n.setObject(hostname);
 
+      DateTime et;
+      try {
+        // "Apr 13 xx:xx:xx"
+        et = DateTime.parse(authTimestamp, DateTimeFormat.forPattern("MMM dd HH:mm:ss"));
```
ameihm0912: Could we move this out of `OpenSSH` into a static function in the parser to generically parse syslog timestamps into `DateTime`?

src/main/java/com/mozilla/secops/parser/OpenSSH.java
```diff
@@ -54,16 +59,27 @@ public OpenSSH(String input, Event e, ParserState state) {
...
+      } catch (IllegalArgumentException exc) {
+        // "Feb  8 xx:xx:xx"
+        et = DateTime.parse(authTimestamp, DateTimeFormat.forPattern("MMM  d HH:mm:ss"));
+      }
+      e.setTimestamp(et.withYear(e.getTimestamp().year().get()));
```
ameihm0912: Yeah not sure what the best way to handle this kind of scenario is. Perhaps if we move the parsing code into a static function as mentioned previously, we can also add some sort of logic in there to correct the year value if it's an unreasonable distance in the future.

Some tests around this would be good to add as well.

src/main/java/com/mozilla/secops/parser/Parser.java
```diff
@@ -140,6 +140,14 @@ private String stripStackdriverEncapsulation(Event e, String input, ParserState
 
       e.setStackdriverProject(getStackdriverProject(entry));
       e.setStackdriverLabels(entry.getLabels());
+      if (entry.getTimestamp() != null) {
+        try {
+          e.setTimestamp(Parser.parseISO8601(entry.getTimestamp()));
+        } catch (IllegalArgumentException exc) {
```
ameihm0912: We should probably catch `IllegalArgumentException` in `parseISO8601` so the function behaves the way it is documented and returns null.

We have the same issue with overwriting with null here as well though, it would be good to try to parse it first and then apply it if we are successful.

src/main/java/com/mozilla/secops/parser/Parser.java
ameihm0912: I think we can just remove this if we address the review feedback, so we never overwrite unless we are confident the resulting parsed timestamp is accurate.

ajvb:
> I added a review comment regarding making `event_timestamp` a mandatory field; that would require addressing the missing event types though and a bunch of changes in the other pipelines, so we can leave that check out for now I think so long as the `AuthProfile` tests validate that timestamp metadata field is being set correctly.

:+1: sounds good.

ameihm0912:
@ajvb looks good, would you mind squashing this into a single commit and we can merge

PR #129, opened 2019-04-19
Load templates from GCS directly

ajvb:
Instead of signing gcs object urls and using the URL template loader,
directly fetch the template and stuff it into a byte array template
loader.

This fixes the problem we were seeing where sometimes urls were not able
to get signed (with no error message as to why) and also improves
runtime performance as templates are all just loaded at class creation.


src/main/java/com/mozilla/secops/authprofile/AuthProfile.java
```diff
@@ -639,8 +640,8 @@ public static Alert createBaseAlert(Event e) {
     a.addMetadata("sourceaddress", n.getSourceAddress());
     a.setCategory("authprofile");
 
-    a.setEmailTemplateName("email/authprofile.ftlh");
-    a.setSlackTemplateName("slack/authprofile.ftlh");
+    a.setEmailTemplateName(TemplateManager.GcsTemplate.AUTHPROFILE_EMAIL.getPath());
```
ameihm0912: Couple comments here,

For the `enum`, should we use something like `Template` instead of `GcsTemplate` considering the paths will apply to resource based templates as well?

Also, I'm not sure on the constants being present in `TemplateManager`. Ideally we would be able to update authprofile without also having to make changes in supporting classes like this (like adding a new constant if we add a new template).

Is there a way we can do this without needing to hardcode pipeline specific templates in `TemplateManager`?

ajvb: > For the `enum`, should we use something like `Template` instead of `GcsTemplate` considering the paths will apply to resource based templates as well?

Yeah, that makes more sense.
 
> Also, I'm not sure on the constants being present in `TemplateManager`. Ideally we would be able to update authprofile without also having to make changes in supporting classes like this (like adding a new constant if we add a new template).
>
> Is there a way we can do this without needing to hardcode pipeline specific templates in `TemplateManager`?

I'm torn here. We are seeing a lot of strings rather than variables (more so in alert metadata than with template names).

In this design, we are able to rely on variables for template names (which could additionally allow for things like checking that all templates exist at start-up) and it's very easy to add a new one. Just set the enum/constant in `TemplateManager` and you are done.

As well, this design allows for a simpler template loading implementation (though implementing dynamic fetching from GCS is _that_ much more code).  Could extend `ByteArrayLoader` to find templates by checking GCS.

PR #135, opened 2019-04-26
Prepend "duo_" onto Duopull's Normalized.object

ajvb:
Mainly to help with the alert message in a simple way. At the moment,
the alert will read "An authentication event for user ajvb was
detected to access admin_login" which is too vague.

### DIFF
--- a/src/main/java/com/mozilla/secops/parser/Duopull.java
+++ b/src/main/java/com/mozilla/secops/parser/Duopull.java
@@ -16,6 +16,8 @@ import org.joda.time.DateTime;
 public class Duopull extends PayloadBase implements Serializable {
   private static final long serialVersionUID = 1L;
 
+  private static final String DUO_OBJECT = "duo";
+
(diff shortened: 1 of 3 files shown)


src/main/java/com/mozilla/secops/parser/Duopull.java
```diff
@@ -88,7 +90,7 @@ public Duopull(String input, Event e, ParserState state) {
         Normalized n = e.getNormalized();
         n.addType(Normalized.Type.AUTH);
         n.setSubjectUser(user);
-        n.setObject(duoPullData.getEventAction());
+        n.setObject(String.format("duo_%s", duoPullData.getEventAction()));
```
ameihm0912: This is always guaranteed to be non-null (`getEventAction()`)?

ajvb: Yeah, github's diff folding is blocking it but this block is wrapped in a `if duoPullData.getEventAction() != null` check.

PR #142, opened 2019-05-08
Set exclude from indexes for state value in Datastore interface

ajvb:


ajvb:
Confirmed working in staging.

PR #145, opened 2019-05-10
Include GeoIP metadata in alerts in AlertFormatter

### LINKED ISSUE #100: Populate geolocation metadata for http_request alerts in bigquery
Geolocation metadata could be useful for running analysis of `http_request` alerts.

ajvb:
* Adds support for Maxmind's ISP db in GeoIP
* AlertFormatter now adds GeoIP metadata to alerts
* Add test for AlertFormatter

Can look up Alert with `id = "4f408b09-4769-41ae-a0a2-85b0fe1187b7"` in nonprod bigquery to look at an example.

Fixes #100

### DIFF
--- a/src/main/java/com/mozilla/secops/OutputOptions.java
+++ b/src/main/java/com/mozilla/secops/OutputOptions.java
@@ -126,6 +126,12 @@ public interface OutputOptions extends PipelineOptions, GcpOptions {
   @Hidden
   void setOutputAlertTemplates(String[] value);
 
+  @Description(
+      "Adds GeoIP metadata to all fields specified (multiple allowed); keys names for alert metadata with an IP address, for example: sourceaddress")
+  String[] getAlertAddressFields();
+
+  void setAlertAddressFields(String[] value);
+
   public static PTransform<PCollection<String>, PDone> compositeOutput(OutputOptions o) {
     return CompositeOutput.withOptions(o);
   }
--- a/src/main/java/com/mozilla/secops/alert/AlertFormatter.java
+++ b/src/main/java/com/mozilla/secops/alert/AlertFormatter.java
@@ -1,6 +1,9 @@
 package com.mozilla.secops.alert;
 
-import com.mozilla.secops.OutputOptions;
+import com.maxmind.geoip2.model.CityResponse;
+import com.maxmind.geoip2.model.IspResponse;
+import com.mozilla.secops.IOOptions;
+import com.mozilla.secops.parser.GeoIP;
 import org.apache.beam.sdk.transforms.DoFn;
 
 /**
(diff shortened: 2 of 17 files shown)


src/main/java/com/mozilla/secops/OutputOptions.java
```diff
@@ -126,6 +126,13 @@
   @Hidden
   void setOutputAlertTemplates(String[] value);
 
+  @Description(
+      "Comma seperated list of alert metadata field names for ip addresses; Defaults to `sourceaddress`")
+  @Default.String("sourceaddress")
```
ameihm0912: Since this modifies the alert format outside of the context of the pipeline, I am wondering if we should **not** have a default for this so it's explicit/obvious that this will occur only if the option is set?

src/main/java/com/mozilla/secops/OutputOptions.java
```diff
@@ -126,6 +126,13 @@
...
+  String getAlertAddressFields();
+
+  void setAlertAddressFields(String value);
```
ameihm0912: Would it be better to use `String[]` here?

```java
void setAlertAddressField(String[] value);
```

```
--alertAddressField=sourceaddress
--alertAddressField=destaddress
```

Just to avoid the extra code to split it?

src/main/java/com/mozilla/secops/alert/AlertFormatter.java
```diff
@@ -13,15 +16,65 @@
...
+            a.addMetadata(addressField + "_as_org", ir.getAutonomousSystemOrganization());
+          }
+        }
+      }
+    }
+  }
+
+  @Setup
+  public void setup() {
+    geoip = null;
```
ameihm0912: I don't think we would need this extra line since it would be unset.

src/main/java/com/mozilla/secops/alert/AlertFormatter.java
```diff
@@ -13,15 +16,65 @@
...
+    if (geoip == null) {
+      return;
+    }
+
+    for (String addressField : addressFields) {
+      String address = a.getMetadataValue(addressField);
+      if (address != null) {
+        if (a.getMetadataValue(addressField + "_city") == null
+            && a.getMetadataValue(addressField + "_country") == null) {
+          if (geoip == null) {}
```
ameihm0912: Not sure what this is here for?

src/main/java/com/mozilla/secops/parser/GeoIP.java
```diff
@@ -20,49 +22,82 @@
    * @param ip IP address string
    * @return MaxmindDB {@link CityResponse}, or null on failure
    */
-  public CityResponse lookup(String ip) {
-    if (!initialized.get()) {
+  public CityResponse lookupCity(String ip) {
+    if (!initialized.get() || geoipCityDb == null) {
```
ameihm0912: Do we need/want the extra null check in here? This should be gated with the atomic read to avoid needing to try to dereference `geoIpCityDb`.

src/main/java/com/mozilla/secops/parser/GeoIP.java
```diff
@@ -20,49 +22,82 @@
...
-  private static synchronized void initialize(String path) throws IOException {
-    InputStream in;
+  /**
+   * Lookup ISP from IP address string
+   *
+   * @param ip IP address string
+   * @return MaxmindDB {@link IspResponse}, or null on failure
+   */
+  public IspResponse lookupIsp(String ip) {
+    if (!initialized.get() || geoipIspDb == null) {
```
ameihm0912: Similar comment to previous around the null check here.

If the intent is to support adding one but not the other, we should probably just have two different atomic flags (one for each).

src/test/java/com/mozilla/secops/alert/TestAlertFormatter.java
```diff
@@ -0,0 +1,72 @@
...
+  public void runFormatter() {
+    IOOptions options = PipelineOptionsFactory.as(IOOptions.class);
+    PCollection<String> res = getTestInput(p).apply(ParDo.of(new AlertFormatter(options)));
+
+    PAssert.that(res)
+        .satisfies(
+            results -> {
+              for (String s : results) {
+                Alert a = Alert.fromJSON(s);
+                assertNotNull(a);
```
ameihm0912: Should we verify the fields are correct here in the no options case?

```java
assertEquals("216.160.83.56", a.getMetadataValue("sourceaddress"));
assertNull(a.getMetadataValue("sourceaddress_city"));
assertNull(a.getMetadataValue("sourceaddress_country"));
```

src/test/java/com/mozilla/secops/authprofile/TestAuthProfile.java
```diff
@@ -199,6 +200,8 @@ public void analyzeTest() throws Exception {
               return null;
             });
 
+    PCollection<String> formatted = res.apply(ParDo.of(new AlertFormatter(options)));
```
ameihm0912: Do we need this? If we aren't going to do anything with the output was thinking we just remove this since it's being tested in another test case.

src/main/java/com/mozilla/secops/OutputOptions.java
```diff
@@ -126,6 +126,12 @@
   @Hidden
   void setOutputAlertTemplates(String[] value);
 
+  @Description(
+      "Alert metadata field names for IP addresses. Enables adding additional GeoIP to all fields specified.")
```
ameihm0912: Minor, but for options that support repeated values (e.g., array options like this) we have `(multiple allowed)` in the description at the end

Also we might want to change it to `getAlertAddressField` but I think it's fine as is too.

PR #162, opened 2019-06-24
AMO pipeline improvements

ameihm0912:


adrianosela:
Looks good so far 👍

src/main/java/com/mozilla/secops/IprepdIO.java
```diff
@@ -30,6 +35,182 @@
...
+     */
+    @JsonProperty("type")
+    public String getType() {
+      return type;
+    }
+
+    /**
+     * Set type value
+     *
+     * @param type Type string
```
adrianosela: "type Type string" ?

src/main/java/com/mozilla/secops/amo/FxaAccountAbuseNewVersion.java
```diff
@@ -60,97 +49,45 @@ public FxaAccountAbuseNewVersion(String monitoredResource) {
...
                     AmoDocker d = e.getPayload();
                     if (d.getEventType().equals(AmoDocker.EventType.GOTPROFILE)) {
-                      if (s.suspectedAccounts.contains(d.getFxaEmail())) {
-                        s.suspectedAddress.add(d.getRemoteIp());
-                        state.write(s);
-
-                        // Also create an alert for the event
+                      // This was a profile fetch, compare the email account against stored account
+                      // reputation information
+                      Integer rep = iprepdReader.getReputation("email", d.getFxaEmail());
```
ajvb: This is sweet. One thought for a future change, it might be nice to have metrics / alerting on how many event's are going through this step to make sure we don't dos iprepd. Looking at prod, it looks like it's very few events at the moment so not a current worry imo.

PR #166, opened 2019-06-28
Refactor ETDFinding subclasses + override equals and hashcode

adrianosela:
* Moving all the inner classes inside the EventThreatDetectionFinding object out to different files for readability
* Override equals and hashcode for etd object + inner classes to allow us to construct PCollection<EventThreatDetectionFinding> without having to deal with the warning message:

```
WARNING: Coder of type class org.apache.beam.sdk.coders.SerializableCoder has a #structuralValue method which does not return true when the encoding of the elements is equal. Element com.mozilla.secops.parser.models.etd.EventThreatDetectionFinding@4704aed
```

### DIFF
--- a/src/main/java/com/mozilla/secops/parser/models/etd/DetectionCategory.java
+++ b/src/main/java/com/mozilla/secops/parser/models/etd/DetectionCategory.java
@@ -1,0 +1,59 @@
+package com.mozilla.secops.parser.models.etd;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.io.Serializable;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class DetectionCategory implements Serializable {
+  private static final long serialVersionUID = 1L;
+
+  private String indicator;
+  private String ruleName;
+  private String technique;
+
+  /**
+   * Get indicator
+   *
+   * @return String
+   */
+  @JsonProperty("indicator")
+  public String getIndicator() {
+    return indicator;
+  }
+
+  /**
+   * Get rule name which triggered finding
+   *
+   * @return String
+   */
+  @JsonProperty("ruleName")
+  public String getRuleName() {
+    return ruleName;
+  }
+
+  /**
+   * Get bad-actor's suspected technique, i.e. "Malware", "Bruteforce", etc...
+   *
+   * @return String
+   */
+  @JsonProperty("technique")
+  public String getTechnique() {
+    return technique;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    DetectionCategory dc = (DetectionCategory) o;
+    return dc.getIndicator().equals(indicator)
+        && dc.getRuleName().equals(ruleName)
+        && dc.getTechnique().equals(technique);
+  }
+
+  @Override
+  public int hashCode() {
+    return indicator.hashCode();
+  }
+
+  public DetectionCategory() {}
+}
--- a/src/main/java/com/mozilla/secops/parser/models/etd/EventThreatDetectionFinding.java
+++ b/src/main/java/com/mozilla/secops/parser/models/etd/EventThreatDetectionFinding.java
@@ -10,195 +10,6 @@ import java.util.ArrayList;
 public class EventThreatDetectionFinding implements Serializable {
   private static final long serialVersionUID = 1L;
 
-  @JsonIgnoreProperties(ignoreUnknown = true)
-  public class DetectionCategory implements Serializable {
-    private static final long serialVersionUID = 1L;
-
-    private String indicator;
-    private String ruleName;
-    private String technique;
-
-    /**
-     * Get indicator
-     *
-     * @return String
-     */
-    @JsonProperty("indicator")
-    public String getIndicator() {
-      return indicator;
-    }
-
-    /**
-     * Get rule name which triggered finding
-     *
-     * @return String
-     */
-    @JsonProperty("ruleName")
-    public String getRuleName() {
-      return ruleName;
-    }
-
-    /**
-     * Get bad-actor's suspected technique, i.e. "Malware", "Bruteforce", etc...
-     *
-     * @return String
-     */
-    @JsonProperty("technique")
-    public String getTechnique() {
-      return technique;
-    }
-
-    public DetectionCategory() {}
-  }
-
-  @JsonIgnoreProperties(ignoreUnknown = true)
-  public static class Evidence implements Serializable {
-    private static final long serialVersionUID = 1L;
-
-    @JsonIgnoreProperties(ignoreUnknown = true)
-    public class SourceLogId implements Serializable {
-      private static final long serialVersionUID = 1L;
-
-      private String insertId;
-      private String timestamp;
-
-      /**
-       * Get insert id
-       *
-       * @return String
-       */
-      @JsonProperty("insertId")
-      public String getInsertId() {
-        return insertId;
-      }
-
-      /**
-       * Get timestamp
-       *
-       * @return String
-       */
-      @JsonProperty("timestamp")
-      public String getTimestamp() {
-        return timestamp;
-      }
-
-      public SourceLogId() {}
-    }
-
-    private SourceLogId sourceLogId;
-
-    @JsonProperty("sourceLogId")
-    public SourceLogId getSourceLogId() {
-      return sourceLogId;
-    }
-
-    public Evidence() {}
-  }
-
-  @JsonIgnoreProperties(ignoreUnknown = true)
-  public class Properties implements Serializable {
-    private static final long serialVersionUID = 1L;
-
-    private String ip;
-    private String location;
-    private String project_id;
-    private String subnetwork_id;
-    private String subnetwork_name;
-    private ArrayList<String> domain;
-
-    /**
-     * Get IP
-     *
-     * @return String
-     */
-    @JsonProperty("ip")
-    public String getIp() {
-      return ip;
-    }
-
-    /**
-     * Get GCP location (analogous to AWS region)
-     *
-     * @return String
-     */
-    @JsonProperty("location")
-    public String getLocation() {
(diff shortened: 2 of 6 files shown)


src/main/java/com/mozilla/secops/parser/models/etd/Evidence.java
```diff
@@ -0,0 +1,71 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.io.Serializable;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class Evidence implements Serializable {
+  private static final long serialVersionUID = 1L;
+
+  @JsonIgnoreProperties(ignoreUnknown = true)
+  public class SourceLogId implements Serializable {
```
ajvb: Why not move this to it's own class as well?

adrianosela: Since evidence has one and only one field (SourceLogId), I figured it made sense to keep that one as an inner class

src/main/java/com/mozilla/secops/parser/models/etd/SourceId.java
```diff
@@ -0,0 +1,47 @@
+package com.mozilla.secops.parser.models.etd;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.io.Serializable;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
```
ajvb: Do you need all of these `@JsonIgnoreProperties(ignoreUnknown = true)`?

src/main/java/com/mozilla/secops/parser/models/etd/Evidence.java
ajvb: Right, but that's true with DetectionCategory, Properties, and SourceId as well. I understood this change having more to do with serialization?

src/main/java/com/mozilla/secops/parser/models/etd/SourceId.java
adrianosela: Not necessarily, but the format of findings vary slightly from finding to finding depending on the rule that triggered it. Having the ignoreUnkown tag allows the parser to handle cases where the finding comes with an unrecognized json field that we havent encountered yet. Im happy to remove all/most and re-add the tag -- or add the new field -- when we encounter it in the future and it breaks the parser.

src/main/java/com/mozilla/secops/parser/models/etd/Evidence.java
ajvb: I do also agree with your mention of readability, in which case I would add that it might be worth moving this to it's own file for consistency.

src/main/java/com/mozilla/secops/parser/models/etd/SourceId.java
ajvb: I hear you, I'd say it's fine to roll with this for now.

src/main/java/com/mozilla/secops/parser/models/etd/Evidence.java
adrianosela: ✅ took it out to a separate file. Will squash commits.

src/main/java/com/mozilla/secops/parser/models/etd/DetectionCategory.java
```diff
@@ -0,0 +1,59 @@
...
+   *
+   * @return String
+   */
+  @JsonProperty("technique")
+  public String getTechnique() {
+    return technique;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
ameihm0912: Is this support class actually passed around in the collection? If not, you don't need to override here as it will be handled by the encapsulating class, we can probably remove this.

src/main/java/com/mozilla/secops/parser/models/etd/EventThreatDetectionFinding.java
```diff
@@ -266,5 +77,16 @@ public SourceId getSourceId() {
     return sourceId;
   }
 
+  @Override
+  public boolean equals(Object o) {
+    EventThreatDetectionFinding etdf = (EventThreatDetectionFinding) o;
+    return etdf.getEventTime().equals(eventTime) && (etdf.getEvidence().equals(evidence));
```
ameihm0912: This needs to be changed; this is going to result in events being dropped that occur at the same time of the same class.

src/main/java/com/mozilla/secops/parser/models/etd/Evidence.java
```diff
@@ -0,0 +1,30 @@
...
+
+  private SourceLogId sourceLogId;
+
+  @JsonProperty("sourceLogId")
+  public SourceLogId getSourceLogId() {
+    return sourceLogId;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
ameihm0912: Can possibly be removed per first comment

src/main/java/com/mozilla/secops/parser/models/etd/Properties.java
```diff
@@ -0,0 +1,97 @@
...
+   *
+   * @return ArrayList<String>
+   */
+  @JsonProperty("domain")
+  public ArrayList<String> getDomain() {
+    return domain;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
ameihm0912: Can possibly be removed per first comment

src/main/java/com/mozilla/secops/parser/models/etd/SourceId.java
```diff
@@ -0,0 +1,47 @@
...
+   *
+   * @return String
+   */
+  @JsonProperty("projectId")
+  public String getProjectId() {
+    return projectId;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
ameihm0912: Can possibly be removed per first comment

src/main/java/com/mozilla/secops/parser/models/etd/SourceLogId.java
```diff
@@ -0,0 +1,46 @@
...
+   *
+   * @return String
+   */
+  @JsonProperty("timestamp")
+  public String getTimestamp() {
+    return timestamp;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
ameihm0912: Can possibly be removed per first comment

src/main/java/com/mozilla/secops/parser/models/etd/EventThreatDetectionFinding.java
adrianosela: ```.getEvidence().equals(evidence)``` will only return true iff the source stack driver log id is the same

src/main/java/com/mozilla/secops/parser/models/etd/SourceId.java
adrianosela: ✅

src/main/java/com/mozilla/secops/parser/models/etd/Properties.java
adrianosela: ✅

src/main/java/com/mozilla/secops/parser/models/etd/DetectionCategory.java
adrianosela: ✅

src/main/java/com/mozilla/secops/parser/models/etd/Evidence.java
adrianosela: ❌ cannot be removed -- gets passed around in the collection

src/main/java/com/mozilla/secops/parser/models/etd/SourceLogId.java
adrianosela: ❌ cannot be removed -- overriden equals is used to compute equals in the ```Evidence``` class

src/main/java/com/mozilla/secops/parser/models/etd/EventThreatDetectionFinding.java
adrianosela: ^^ this claim is being tested in https://github.com/mozilla-services/foxsec-pipeline/pull/167

ameihm0912: Is this log ID the ID of this particular event itself that is guaranteed to be unique?

I assumed it being an array it was potentially the source log ID messages that resulted in the finding being generated. If that's the case we wouldn't want to use it as an ID value.

If it is a unique event ID it seems odd it would be an array.

adrianosela: @ameihm0912  you are absolutely correct.... :(  what uniquely identifies an etd finding is the "insertId" of the finding itself. Which at this point in the code has been stripped along the rest of the stackdriver wrapper. Will have to figure out a way to embed the id of the finding log itself within the EventThreatDetection object and use that in the .equals function. I will update #167 to reflect this by next week.

Nice catch.

adrianosela: Edit: + I had assumed that it was impossible for a set of "evidence" (source logs) to generate more than one finding, which may or may not be the case. -- but the stackdriver log of the finding itself is guaranteed to be unique, so lets go with that.

ameihm0912: @adrianosela no worries; for now I have just reverted the changes made to master in #166 until it is ready to go into that branch.

Let's start with a new branch/PR with the changes you need here so we can validate them going in at once rather then piecing out fixing #166 up.

adrianosela:
Follow up PR: https://github.com/mozilla-services/foxsec-pipeline/pull/168

PR #167, opened 2019-06-28
[DO NOT MERGE] Remove unnecessary .equals and .hashCode from ETD model classes

adrianosela:
As noted by :alm in https://github.com/mozilla-services/foxsec-pipeline/pull/166, most of the supporting classes for the EventThreatDetectionFinding model do not need an overridden .equals and .hashCode. 

This PR removes those funcs.


ameihm0912:
See comments in #166, will be included with the change set this diff applies to in a follow up PR

PR #168, opened 2019-06-30
Event Threat Detection Model Refactor

adrianosela:
* Moving inner classes of the ETD JSON model out to separate files for readability


src/main/java/com/mozilla/secops/parser/models/etd/EventThreatDetectionFinding.java
```diff
@@ -266,5 +97,30 @@ public SourceId getSourceId() {
     return sourceId;
   }
 
+  @Override
+  public boolean equals(Object o) {
```
ajvb: imo, given your comment below, I'd say we just remove this `equals` function for now. If we get duplicate alerts, we can come back to this.

adrianosela: sounds good to me. The warning logs can get quite spammy but I guess we dont see those at all when running in dataflow(?) -- will wait on @ameihm0912 's input before removing.

PR #170, opened 2019-07-03
Implement Geo Velocity in AuthProfile

ajvb:
Adds geo velocity anomaly detection within authprofile.

Also includes:
* changing StateModel so that old entries will stay around for 30 days by extending prune age and adding an `isExpired()` that is at 10 days
* Removes all geoip stuff from the parsers and puts it into Normalized
* Adds lat/lon to StateModel and Normalized


src/main/java/com/mozilla/secops/authprofile/StateModel.java
```diff
@@ -16,15 +18,57 @@
  * <p>Used by {@link AuthProfile}.
  */
 public class StateModel {
-  private final Long DEFAULTPRUNEAGE = 864000L; // 10 days
+  private final Long DEFAULTPRUNEAGE = 864000L * 3; // 30 days
```
ameihm0912: Should this just be declared static too?

src/main/java/com/mozilla/secops/authprofile/StateModel.java
```diff
@@ -16,15 +18,57 @@
...
+     */
+    @JsonProperty("latitude")
+    public Double getLatitude() {
+      return latitude;
+    }
+
+    /**
+     * Set model longitude field
+     *
+     * @param longitutde longitude double value
```
ameihm0912: typo

src/main/java/com/mozilla/secops/authprofile/StateModel.java
```diff
@@ -16,15 +18,57 @@
...
+  @JsonIgnoreProperties(ignoreUnknown = true)
   static class ModelEntry {
+    private Double longitude;
+    private Double latitude;
     private DateTime timestamp;
 
+    /**
+     * Set model latitude field
+     *
+     * @param latitutde Latitude double value
```
ameihm0912: typo

src/main/java/com/mozilla/secops/parser/BmoAudit.java
```diff
@@ -23,8 +22,6 @@
 
   private String msg;
   private String remoteIp;
-  private String remoteIpCity;
```
ameihm0912: My pref here would be for us to not remove these fields from the given event payload types. It should be up to the event to decide to set the normalized field values if it wants to, but I don't think pipelines should be required to access normalized fields to get this type of information.

Is there any way we can adjust this PR so we aren't changing all of the events in it for now and then later down the road if we need this we can look into it?

src/main/java/com/mozilla/secops/parser/Cloudtrail.java
```diff
@@ -25,9 +24,6 @@
 
   private CloudtrailEvent event;
 
-  private String sourceAddressCity;
```
ameihm0912: Same comment as previous for this one

src/main/java/com/mozilla/secops/parser/Duopull.java
```diff
@@ -102,16 +101,6 @@ public Duopull(String input, Event e, ParserState state) {
             n.setSubjectUserIdentity(resId);
           }
         }
-
-        if (sourceAddress != null) {
```
ameihm0912: Same comment as previous for this one; maybe here we should adjust this so it sets the required fields both internally in the class and also in the normalized component.

src/main/java/com/mozilla/secops/parser/GcpAudit.java
```diff
@@ -26,8 +25,6 @@
   private String principalEmail;
   private String resource;
   private String callerIp;
-  private String callerIpCity;
```
ameihm0912: Same comment here

src/main/java/com/mozilla/secops/parser/Normalized.java
```diff
@@ -123,6 +126,24 @@ public void setSourceAddress(String addr) {
     sourceAddress = addr;
   }
 
+  /**
+   * Set source address field and attempt to set geo IP metadata.
+   *
+   * @param addr Source address
+   */
+  public void setSourceAddressWithGeoIPMetadata(String addr, Parser parser) {
```
ameihm0912: To me the concept seems sound here but as mentioned, the normalized data set is intended to just be a reflection of elements that are already present in the event itself, and it shouldn't be the only source of this information.

src/main/java/com/mozilla/secops/parser/OpenSSH.java
```diff
@@ -24,8 +23,6 @@
   private String user;
   private String authMethod;
   private String sourceAddress;
-  private String sourceAddressCity;
-  private String sourceAddressCountry;
```
ameihm0912: Same comment here

src/main/java/com/mozilla/secops/parser/Taskcluster.java
```diff
@@ -14,8 +13,6 @@
   private com.mozilla.secops.parser.models.taskcluster.Taskcluster data;
   private String subject;
   private String sourceAddress;
-  private String sourceAddressCity;
-  private String sourceAddressCountry;
```
ameihm0912: Same comment here

ajvb:
@ameihm0912 Merged in #172 (and fixed the other issues pointed out)

src/main/java/com/mozilla/secops/GeoUtil.java
```diff
@@ -0,0 +1,51 @@
+package com.mozilla.secops;
+
+/** Geo math utilities */
+public class GeoUtil {
+
+  /**
+   * haversin(0)
```
adrianosela: looks good, but is there a reason we roll our own math funcs? https://github.com/locationtech/spatial4j

ameihm0912: since it's a pretty simple function no reason to introduce a dependency for it.

src/main/java/com/mozilla/secops/parser/Normalized.java
```diff
@@ -125,6 +126,24 @@ public void setSourceAddress(String addr) {
...
```
ameihm0912: Do we still need this function? I don't think this is required anymore right?

src/test/java/com/mozilla/secops/authprofile/TestAuthProfile.java
```diff
@@ -404,4 +406,77 @@ public void analyzeNamedSubnetsTest() throws Exception {
...
+
+    try {
+      PCollection<String> inputOne =
+          TestUtil.getTestInput("/testdata/authprof_geovelocity_buffer1_part1.txt", p1);
+      PCollection<String> inputTwo =
+          TestUtil.getTestInput("/testdata/authprof_geovelocity_buffer1_part2.txt", p2);
+      PCollection<String> inputThree =
+          TestUtil.getTestInput("/testdata/authprof_geovelocity_buffer1_part3.txt", p3);
+
+      PCollection<Alert> resultsOne = AuthProfile.processInput(inputOne, options);
```
ameihm0912: Are these just three different test cases? It doesn't look like `resultsOne` is checked at all anywhere?

src/test/java/com/mozilla/secops/parser/ParserIdentityTest.java
```diff
@@ -33,14 +33,14 @@ public void testOpenSSHStackdriverWithIdentity() throws Exception {
     assertEquals("riker", o.getUser());
     assertEquals("publickey", o.getAuthMethod());
     assertEquals("127.0.0.1", o.getSourceAddress());
-    assertNull(o.getSourceAddressCity());
-    assertNull(o.getSourceAddressCountry());
```
ameihm0912: With the changes made to how we apply the geo stuff I don't think we need to change these around anymore?

src/test/java/com/mozilla/secops/authprofile/TestAuthProfile.java
ajvb: No, they are split up like this so that the test can manipulate the "current time". Check out the series of `DateTimeUtils.setCurrentMillisFixed()` and `pN.run().waitUntilFinish();` calls at the bottom. `resultsOne` is just there to set the "last location" for `wriker@mozilla.com`

ameihm0912: So does this then adjust what the actual system clock will return when queried? That doesn't really seem ideal if that is the case, would this potentially also impact beam's internals?

Is this required just because we request current time in the pipeline transform and want it to appear as a certain value? Could we not just use the event timestamp instead so this isn't needed?

ajvb: >So does this then adjust what the actual system clock will return when queried?

Yes, but only for Joda's `DateTime` and `DateTimeUtils` calls.

>...would this potentially also impact beam's internals?

Yes, as they use joda's `DateTimeUtils.currentTimeMillis()` in places, but I believe that is what we want here. We want to simulate the pipeline running multiple times over a manner of hours.

>Is this required just because we request current time in the pipeline transform and want it to appear as a certain value?

Not just for that reason, but also because our use of joda's `DateTime()` in `StateModel` (https://github.com/mozilla-services/foxsec-pipeline/blob/master/src/main/java/com/mozilla/secops/authprofile/StateModel.java#L81-L83).

Since the model entries timestamp is set like this, it makes the most sense to calculate the time difference as so:
```
Long timeDifference = (DateTimeUtils.currentTimeMillis() / 1000) - (lastLocation.getTimestamp().getMillis() / 1000);
```
(https://github.com/mozilla-services/foxsec-pipeline/pull/170/files#diff-e2da9f9c753def490ec4bba2c9bfed47R598)


I could definitely be missing something, but in this test case we are not testing windowing or anything related to beam having a view on what time really is. Another way to do this could be supporting passing in a distinct mock clock that is used everywhere we call `joda.[DateTime|DateTimeUtils]`, but that would also bring the test further away from how this code would run in a pipeline

ameihm0912: > Not just for that reason, but also because our use of joda's `DateTime()` in `StateModel`

Ah ok, yeah I see what you mean now. Yeah that is problematic.

As another option, what about splitting the tests up into a few types and doing something like this:

One would set `maxKilometersPerSecond` to something very low, so that should trigger as long as the long/lat are different, that could test the end to end initial creation and subsequent evaluation of the state information even if the event timestamps were very close together.

Then as a second test, preload datastore with mock data, and evaluate that an alert is generated based on feeding in a new event with the default pipeline configuration?

We could possibly even have a pipeline configuration option that would use the event timestamp with `StateModel` instead, as `updateEntry` already has an overridden function signature that supports passing in a timestamp instead of using current time. The test could just set this pipeline option and then we will use event timestamp with state model?

Given all the dependencies that make use of Joda I feel like we should come up with a better way to test this deterministically that isn't going to potentially cause problems in the future.

ajvb: Reworked it without messing with the time in https://github.com/mozilla-services/foxsec-pipeline/pull/170/commits/d6fce736f9cf4d1f42b2c4aa7f2c3adadf3cfc2d

ameihm0912:
@ajvb looks like this one has a merge conflict now

ajvb:
@ameihm0912 rebased off master

PR #171, opened 2019-07-04
updates to amo pipeline

ameihm0912:


src/main/java/com/mozilla/secops/amo/Amo.java
```diff
@@ -63,11 +68,66 @@
...
+                    options.getAddonMatchCriteria())));
+
     return resultsList.apply("amo flatten output", Flatten.<Alert>pCollections());
   }
 
   /** Runtime options for {@link Customs} pipeline. */
-  public interface AmoOptions extends PipelineOptions, IOOptions {}
+  public interface AmoOptions extends PipelineOptions, IOOptions {
+    @Description("On login if account matches regex, generate alert regardless of reputation")
+    String[] getAccountMatchBanOnLogin();
```
ajvb: If we are going to be making frequent updates to this list, could it be worth moving to something like datastore?

ameihm0912: yeah I think eventually, or possibly even in iprepd 👍

PR #172, opened 2019-07-04
SourcePayloadBase, unify GeoIP processing

ameihm0912:
Unifies processing of source IP GeoIP fields in extended payload base
class; this ensures the GeoIP fields are available both from the payload
type and also normalized fields.


ameihm0912:
@ajvb re #170 what do you think of doing something like this?

This is similar to what you were doing with the normalized fields except the information is also available from the base payload type too and keeps the existing payload methods intact.

I think this should work for what you want to do but if not we can try another method too.

PR #173, opened 2019-07-08
AMO stackdriver test

ameihm0912:
Add a test to verify AMO SD parsing.

Also fix a couple minor issues in the AMO parser here.


ameihm0912:
> Looks likes travis is being weird / there is a `ConcurrentModification` exception on TestAmo.amoFxaAbuseNewVersionTest

Yeah I don't think this is a travis issue, it also pops up from time to time just running tests locally. I am not sure what the cause of it is yet.

PR #169, opened 2019-07-01
Gatekeeper Pipeline

adrianosela:
### AWS GD + GCP ETD Pipeline which does the following:

- Input stage using our standard composite input transforms
- Use GlobalTriggers and a 60 second trigger interval to window the event stream
- Split this PCollection into two collections, one containing GD and one ETD
- Create a new ETD analysis transform that will be responsible for filtering out events we don't care about, and creating Alert objects for the ones we do care about
- Same for GD
- Flatten these collections into one collection
- Output using our existing composite output transforms

PR #174, opened 2019-07-09
Gatekeeper Pipeline

adrianosela:
Re-PR of https://github.com/mozilla-services/foxsec-pipeline/pull/169 for cleanliness.

### AWS GD + GCP ETD Pipeline which does the following:

- Input stage using our standard composite input transforms
- Use GlobalTriggers and a 60 second trigger interval to window the event stream
- Split this PCollection into two collections, one containing GD and one ETD
- Create a new ETD analysis transform that will be responsible for filtering out events we don't care about, and creating Alert objects for the ones we do care about
- Same for GD
- Flatten these collections into one collection
- Output using our existing composite output transforms


ameihm0912:
@adrianosela I created #175 which we will want to add here as well, but we can do that in a subsequent PR.

PR #181, opened 2019-07-18
add amo pipeline updates

ameihm0912:


src/main/java/com/mozilla/secops/amo/FxaAccountAbuseAlias.java
```diff
@@ -23,7 +23,10 @@
 /**
  * Analysis for aliased account usage
  *
- * <p>A session gap duration of 120 minutes is used with early firings every 60 seconds.
+ * <p>A session gap duration of 120 minutes is used with early firings every 60 seconds. Within the
+ * window accounts in used are normalized and grouped together using the normalized value as a key.
```
ajvb: `Within the window accounts in used` - should that be `in use`?

src/main/java/com/mozilla/secops/amo/FxaAccountAbuseAlias.java
```diff
@@ -23,7 +23,10 @@
...
+ * If the number if in-use aliases for a given account exceeds the configured value, and alert is
```
ajvb: Should that second `if` be `of`?

src/main/java/com/mozilla/secops/amo/Amo.java
```diff
@@ -127,6 +157,63 @@
     Integer getAddonMatchSuppressRecovery();
 
     void setAddonMatchSuppressRecovery(Integer value);
+
+    @Description(
+        "Generate multi match alert if uploads exceed this number in given window with same file name")
```
ajvb: "in a given window"?

src/main/java/com/mozilla/secops/amo/AddonMatcher.java
```diff
@@ -98,12 +99,31 @@ public void processElement(ProcessContext c) {
...
+                          String buf = email;
+                          String nb = MiscUtil.normalizeEmailPlus(email);
+                          if (!email.equals(nb)) {
+                            buf += ", " + nb;
+                          }
+                          nb = MiscUtil.normalizeEmailPlusDotStrip(email);
+                          if (!email.equals(nb)) {
+                            buf += ", " + nb;
+                          }
+                          alert.addMetadata("email", buf);
```
ajvb: not blocking, but packing multiple email addresses into one metadata kv pair seems like it may lead to pain later (and pain now with regards to analyzing and searching through alerts in bq, via queries or in something like redash)

src/main/java/com/mozilla/secops/amo/FxaAccountAbuseAlias.java
ameihm0912: yep, fixed

src/main/java/com/mozilla/secops/amo/AddonMatcher.java
ameihm0912: yeah, completely agree. I'm not sure what the best way to handle this is right now, we wouldn't want something like multiple keys either I suspect.

It's probably worth standardizing on some sort of metadata value array type that applies constraints on how this is constructed.

Issue #79, opened 2019-02-26
Write tests for IprepdIO


ajvb:
https://github.com/mozilla-services/foxsec-pipeline/pull/77#event-2163555968

PR #183, opened 2019-07-19
add tests for IprepdIO, fix malformed URL handling

ameihm0912:
Adds write tests for IprepdIO; we had some read tests in here already so this should cover off what is needed in #79 

Also fixes unhandled exceptions when the constructed URL is malformed


ameihm0912:
@ajvb removing review flag here for now; I think based on that issue we also want whitelist tests, so will add those and request review once it is ready.

PR #185, opened 2019-07-19
Thread Safe Alert Metadata - using CopyOnWriteArrayList<E>

adrianosela:
This seems to have eliminated "ConcurrentModificationException"s

We should find a way to deterministically test this...


adrianosela:
Closing in favor of https://github.com/mozilla-services/foxsec-pipeline/pull/191

PR #186, opened 2019-07-19
Migrate to new whitelisted object datastore entries

### LINKED ISSUE #180: Use iprepd whitelist for other types besides just IP
Currently, the code the checks if an entry has been whitelisted for iprepd only works with IP addresses; this should be expanded so it works with other types as well (e.g., email).

Depends on https://github.com/mozilla-services/foxsec-pipeline-contrib/issues/21

ajvb:
Migrate to new "whitelisted object" concept, changing the class definition and using the type as the datastore kind.

Depends on https://github.com/mozilla-services/foxsec-pipeline-contrib/pull/23

Fixes #180


src/main/java/com/mozilla/secops/IprepdIO.java
```diff
@@ -514,7 +537,7 @@ public void setCreatedBy(String createdBy) {
    * @param a Alert to add metadata to
    */
   public static void addMetadataIfIpWhitelisted(String ip, Alert a) throws IOException {
```
adrianosela: What is the reason behind keeping ```addMetadataIfIpWhitelisted(String, Alert)``` and ```addMetadataIfIpWhitelisted(String, Alert, String)```?

ajvb: Smaller number of changes to the repo, don't need to export `whitelistedIpKind` (or create a new `IP_TYPE` constant or similar) and use it everywhere, and we are not loosing any clarity with making these alias functions

ameihm0912:
@ajvb any chance we can kill that merge commit in this PR too

src/main/java/com/mozilla/secops/IprepdIO.java
```diff
@@ -437,29 +437,49 @@ public static void addMetadataSuppressRecovery(Integer value, Alert a) {
     a.addMetadata(IPREPD_SUPPRESS_RECOVERY, value.toString());
   }
 
-  /** WhitelistedIp contains the metadata associated with a whitelisted ip. */
-  public static class WhitelistedIp {
-    private String ip;
```
ameihm0912: What if we were to keep the `ip` field in the new object class? With the intent being to support reverse compatibility, so we can load both the old format or the new format with the same class. There is likely some way to handle this nicely automatically as part of deserialization after reading the value from Datastore.

We would probably also want `@JsonIgnoreProperties(ignoreUnknown = true)` so the JSON code does not complain when reading the older format records?

src/main/java/com/mozilla/secops/IprepdIO.java
```diff
@@ -502,10 +522,13 @@ public void setCreatedBy(String createdBy) {
   }
 
   /** Kind for whitelisted IP entry in Datastore */
-  public static final String whitelistedIpKind = "whitelisted_ip";
```
ameihm0912: can we leave these in, perhaps as `legacyWhitelistedIpKind` or something, and `legacyWhitelistedIpNamespace`?

src/main/java/com/mozilla/secops/IprepdIO.java
```diff
@@ -529,18 +552,49 @@ public static void addMetadataIfIpWhitelisted(String ip, Alert a) throws IOExcep
...
+   * @param type Type of object (usually "ip" or "email")
+   * @param a Alert to add metadata to
+   * @param datastoreProject If Datastore is in another project, non-null project ID
+   */
+  public static void addMetadataIfObjectWhitelisted(
+      String obj, String type, Alert a, String datastoreProject) throws IOException {
+    if (obj == null || type == null || a == null) {
+      return;
+    }
+
```
ameihm0912: If the type is IP, we should be able to fairly easily make a request against the legacy Datastore namespace/kind in addition to the new format

ajvb:
@ameihm0912 Should be good for a re-review. Added support back for the now "legacy" whitelisted IP entries and included that in the unit test.

ameihm0912:
@ajvb guessing this is OK to have that label removed and be merged now with these changes?

PR #187, opened 2019-07-19
Thread Safe Alert Metadata - use HashMap to avoid iterating in getMetadata(String k)

adrianosela:
Looks like the "Concurrent Modification" which raises ConcurrentModificationException is mitigated by using a map as a structure, as there is no need to iterate through all metadata contents to get a single value out.


adrianosela:
Closing this -- the format of alerts MUST be SQL friendly in order to be stored in BigQuery. Using a java.util Map breaks this requirement.

Issue #175, opened 2019-07-10
Implement escalation metadata and alert suppression in Gatekeeper pipeline


ameihm0912:
#174 adds a pipeline for monitoring output from ETD and GD.

In the current state, the pipeline will generate alerts but it does not currently add any escalation metadata to the alerts that will result in special handling in `AlertIO`.

Example from the authentication pipeline:

https://github.com/mozilla-services/foxsec-pipeline/blob/7a27d9421f9ba44ccdbdd327de4c080690f68bbf/src/main/java/com/mozilla/secops/authprofile/AuthProfile.java#L255-L263

We will want to add a similar notification pipeline option to the gatekeeper pipeline and include this metadata option when we generate an alert if the option is set in the configuration, if the option is not set the pipeline will behave as it currently does.

We will also want to potentially look at hooking `AlertSuppressor` up to the end of the analysis transforms and suppressing possible repeated alerts to avoid generating a large number of escalation notifications under certain circumstances.

adrianosela:
Update: Missing ETD alert suppressor

PR #189, opened 2019-07-22
Gatekeeper: Implement Alert Escalation Metadata

adrianosela:
closes https://github.com/mozilla-services/foxsec-pipeline/issues/175

### DIFF
--- a/src/main/java/com/mozilla/secops/OutputOptions.java
+++ b/src/main/java/com/mozilla/secops/OutputOptions.java
@@ -138,6 +138,12 @@ public interface OutputOptions extends PipelineOptions, GcpOptions {
 
   void setAlertAddressFields(String[] value);
 
+  @Description(
+      "Email address to receive critical alert notifications if application generates them")
+  String getCriticalNotificationEmail();
+
+  void setCriticalNotificationEmail(String value);
+
   public static PTransform<PCollection<String>, PDone> compositeOutput(OutputOptions o) {
     return CompositeOutput.withOptions(o);
   }
--- a/src/main/java/com/mozilla/secops/gatekeeper/GatekeeperOptions.java
+++ b/src/main/java/com/mozilla/secops/gatekeeper/GatekeeperOptions.java
@@ -16,4 +16,20 @@ public interface GatekeeperOptions extends PipelineOptions, IOOptions {
   String[] getIgnoreGDFindingTypeRegex();
 
   void setIgnoreGDFindingTypeRegex(String[] value);
+
+  @Description("Escalate ETD Findings for any finding rules that match regex (multiple allowed)")
+  String[] getEscalateETDFindingRuleRegex();
+
+  void setEscalateETDFindingRuleRegex(String[] value);
+
+  @Description(
+      "Escalate GuardDuty Findings for any finding types that match regex (multiple allowed)")
+  String[] getEscalateGDFindingTypeRegex();
+
+  void setEscalateGDFindingTypeRegex(String[] value);
+
+  @Description("Email address to receive critical alert notifications")
+  String getCriticalNotificationEmail();
+
+  void setCriticalNotificationEmail(String value);
 }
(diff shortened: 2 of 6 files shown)


ajvb:
@adrianosela It would be nice to see some additions to the gatekeeper tests for the escalation regexes and the `notify_direct` metadata.

adrianosela:
Added some tests and 
imaged commits @ajvb

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -73,6 +73,32 @@ public void processElement(ProcessContext c) {
...
+     * static initializer for alert generation / escalation
+     *
+     * @param opts {@link GatekeeperOptions} pipeline options
+     */
+    public GenerateAlerts(GatekeeperOptions opts) {
+      critNotifyEmail = opts.getCriticalNotificationEmail();
+      String[] escalateRegexes = opts.getEscalateGDFindingTypeRegex();
+
+      escalate = new ArrayList<Pattern>();
+      if (escalateRegexes != null) {
```
ajvb: What do you think about the logic of the escalation regexes to `if regexes == null` then automatically set it to `*` (or `.+`)?

src/main/java/com/mozilla/secops/gatekeeper/GatekeeperOptions.java
```diff
@@ -16,4 +16,20 @@
...
+  @Description(
+      "Escalate GuardDuty Findings for any finding types that match regex (multiple allowed)")
+  String[] getEscalateGDFindingTypeRegex();
+
+  void setEscalateGDFindingTypeRegex(String[] value);
+
+  @Description("Email address to receive critical alert notifications")
+  String getCriticalNotificationEmail();
+
+  void setCriticalNotificationEmail(String value);
```
ajvb: Might be worth taking this and https://github.com/mozilla-services/foxsec-pipeline/blob/master/src/main/java/com/mozilla/secops/authprofile/AuthProfile.java#L613-L616 and moving them into https://github.com/mozilla-services/foxsec-pipeline/blob/master/src/main/java/com/mozilla/secops/OutputOptions.java rather than duplicating these options.

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
adrianosela: I think that would be smart 👍 -- will add

adrianosela: ✅

src/main/java/com/mozilla/secops/gatekeeper/GatekeeperOptions.java
adrianosela: ✅  thanks for the suggestion

ameihm0912:
@adrianosela can you rebase this against master? looks like we have some conflicts in here

src/main/java/com/mozilla/secops/OutputOptions.java
```diff
@@ -138,6 +138,11 @@
 
   void setAlertAddressFields(String[] value);
 
+  @Description("Email address to receive critical alert notifications")
```
ameihm0912: Minor, but should we add something like `if application generates them` to the end of this if we are going to make it a global option?

ameihm0912:
@adrianosela also can you squash out the merge commits in here before we merge it too

adrianosela:
> @adrianosela lgtm
> 
> Note that this is missing the alert suppression requirement we had in #175 so we would still need to do that, up to you if you prefer to do that in another PR though (might be easier)

Yeah, all this code has already been looked at so lets just get it in. Will make a follow up PR.

adrianosela:
@ameihm0912 will let you hit the green since I just pushed -f

PR #191, opened 2019-07-23
protect metadata with mutex during serialization

ameihm0912:


ameihm0912:
@ajvb @adrianosela any thoughts on this as an approach? this should eliminate the need for the copies and protects the array with a mutex during serialization.

adrianosela:
How come you done need to hold the lock to read? i.e. getMetadataValue(), and all instances in the code base that use the array after getMetadata()?

ameihm0912:
> How come you dont need to hold the lock to read? i.e. getMetadataValue(), and all instances in the code base that use the array after getMetadata()?

I don't suspect we will have a problem there as in the context of our code we are always reading/writing to metadata from a single thread, that being said though if we run into this again it might be worth looking into adjusting it as you mention.

adrianosela:
ran this ~200 times with no failures. I say its good to merge.

PR #195, opened 2019-07-24
Gatekeeper - Fix null ptr exception

adrianosela:
Without this change, dataflow gives me: 
Screen Shot 2019-07-24 at 3 52 05 PM
Which is strange; there really shouldnt be a null pointer exception here given that I am initializing the escalate array it in the static initializer / constructor for the transform.
```
for (Pattern p : escalate) {
    if (p.matcher(f.getType()).matches()) {
        addEscalationMetadata(a);
        break;
    }
}
```
... but oh well. Thoughts?

### DIFF
--- a/src/main/java/com/mozilla/secops/gatekeeper/ETDTransforms.java
+++ b/src/main/java/com/mozilla/secops/gatekeeper/ETDTransforms.java
@@ -96,7 +96,7 @@ public class ETDTransforms implements Serializable {
 
     private static final String alertCategory = "gatekeeper:gcp";
 
-    private static List<Pattern> escalate;
+    private static List<Pattern> escalate = new ArrayList<Pattern>();
     private static String critNotifyEmail;
 
     /**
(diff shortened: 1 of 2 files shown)


ajvb:
@adrianosela Couple thoughts:

1. It'd be good to include a regression test that triggers this behavior.
2. On another note, from looking at https://github.com/mozilla-services/foxsec-pipeline/pull/189/files - I may be mistaken, but it looks like when commits got squashes, the code around including `.+` in the escalation regexes if none are set got deleted? (e.g. this comment: https://github.com/mozilla-services/foxsec-pipeline/pull/189/files#r306057409)

adrianosela:
> @adrianosela Couple thoughts:
> 
>     1. It'd be good to include a regression test that triggers this behavior.
> 
>     2. On another note, from looking at https://github.com/mozilla-services/foxsec-pipeline/pull/189/files - I may be mistaken, but it looks like when commits got squashes, the code around including `.+` in the escalation regexes if none are set got deleted? (e.g. this comment: https://github.com/mozilla-services/foxsec-pipeline/pull/189/files#r306057409)

1 -- will try to reproduce.
2 -- You are correct, will PR the change

adrianosela:
I was fixing the wrong thing. Closing in favor of https://github.com/mozilla-services/foxsec-pipeline/pull/196, which fixes the NPE by removing "static" declarations the exclude arraylist

PR #196, opened 2019-07-24
Gatekeeper - Implicit escalate all findings + remove bad static declaration

adrianosela:
We want the default alert escalation setting to include all finding types if no alert escalation regexes are provided.

This had already been implemented in https://github.com/mozilla-services/foxsec-pipeline/pull/189, but the change got removed when squashing


ajvb:
You had some tests in #189 - want to include them in this PR as well?

PR #198, opened 2019-07-25
Gatekeeper: GuardDuty Alert Suppressor

adrianosela:
Alert suppressor for GuardDuty findings

Part of https://github.com/mozilla-services/foxsec-pipeline/issues/175


adrianosela:
This one's ready ✅

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -17,6 +18,7 @@
 /** Implements various transforms on AWS GuardDuty {@link Finding} Events */
 public class GuardDutyTransforms implements Serializable {
   private static final long serialVersionUID = 1L;
+  private static final long defaultSuppressionWindow = (long) 60 * 15; // 15 minutes
```
ameihm0912: I think we can just remove this and set a `Default.Long` annotation on the pipeline option

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -164,4 +171,61 @@ public void processElement(ProcessContext c) {
               }));
     }
   }
+
+  /**
+   * Suppress Alerts for repeated GuardDuty Findings.
+   *
+   * <p>A "repeated finding" in GuardDuty means the same (potential) bad actor is performing the
+   * same action against the same resource in your AWS environment. Findings are uniquely identified
+   * by their "id".
```
ameihm0912: Hmm, I'm not too sure if this is what we want here.

As an example, say a scenario with 1000 instances where they all generate a finding. I'm guessing that would be 1000 distinct `id` values? We definitely don't want 1000 alerts with the escalation flag set going out.

Originally I was thinking we'd do something like a key that is a combination of the AWS account ID and the finding type or something along those lines. That would suppress duplicate notifications for the same finding type from the same AWS account which is ideal, but it also would result in some alerts not being inserted into BigQuery.

@ajvb do you have any thoughts on this? Does this potentially require some sort of customized suppression logic so we still generate the alerts but perhaps only a subset have the notification flag set on them?

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -164,4 +171,61 @@ public void processElement(ProcessContext c) {
...
+   * or 6 hours (default).
+   * https://docs.aws.amazon.com/guardduty/latest/ug/guardduty_findings_cloudwatch.html#guardduty_findings_cloudwatch_notification_frequency
+   *
+   * <p>This transform adds a second layer of protection against generation of alerts for repeated
+   * findings
+   */
+  public static class SuppressAlerts extends PTransform<PCollection<Alert>, PCollection<Alert>> {
+    private static final long serialVersionUID = 1L;
+
+    private static Long alertSupressionWindow;
```
ameihm0912: typo here

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -164,4 +171,61 @@ public void processElement(ProcessContext c) {
...
+
+    /**
+     * static initializer for alert suppression
+     *
+     * @param opts {@link Options} pipeline options
+     */
+    public SuppressAlerts(Options opts) {
+      alertSupressionWindow = opts.getAlertSuppressionSeconds();
+      if (alertSupressionWindow == null) {
```
ameihm0912: Can remove this check if we just set a default annotation on the option

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -164,4 +171,61 @@ public void processElement(ProcessContext c) {
...
+      return input
+          .apply(
+              ParDo.of(
+                  new DoFn<Alert, KV<String, Alert>>() {
+                    private static final long serialVersionUID = 1L;
+
+                    @ProcessElement
+                    public void processElement(ProcessContext c) {
+                      Alert a = c.element();
+                      if (a == null) {
```
ameihm0912: I don't think this is possible to occur but the check doesn't hurt

adrianosela:
We can have a small chat about the suppression key being used next week. The ETD stuff is not being added because ETD findings dont seem to have any unique identifier other than the encompassing Stackdriver log id -- We should also talk about that next week.

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
adrianosela: @ameihm0912 I think the findings are very well defined i.e. there wont be an event where two distinct findings are generated unless two distinct actions are being done (i.e. brute force + port scan = 2 distinct findings). However if the same thing is being done twice, it only increments a counter on the finding and a new cloudwatch event is not generated.

adrianosela:
Closing in favor of https://github.com/mozilla-services/foxsec-pipeline/pull/201

PR #199, opened 2019-07-26
cfgtick generation

ameihm0912:
Adds initial support for generation of periodic configuration ticks from
CompositeInput.

The messages are currently limited to current pipeline runtime options,
but are intended to be expanded to allow transforms to describe
functionality as well for inclusion in these messages.

By default, the new functionality is disabled.

The HTTPRequest pipeline has been modified to demonstrate use of the
generator.

### DIFF
--- a/src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
+++ b/src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
@@ -1203,7 +1203,7 @@ public class HTTPRequest implements Serializable {
               events
                   .apply(
                       "cfgtick processor",
-                      ParDo.of(new CfgTickProcessor("httprequest", "category")))
+                      ParDo.of(new CfgTickProcessor("httprequest-cfgtick", "category")))
                   .apply(new GlobalTriggers<Alert>(5)));
     }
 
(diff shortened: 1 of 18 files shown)


src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
```diff
@@ -1183,6 +1196,17 @@ private static void runHTTPRequest(HTTPRequestOptions options) {
...
     }
 
+    // If configuration ticks were enabled, enable the processor here too
+    if (options.getGenerateConfigurationTicksInterval() > 0) {
+      resultsList =
+          resultsList.and(
+              events
+                  .apply(
+                      "cfgtick processor",
+                      ParDo.of(new CfgTickProcessor("httprequest", "category")))
```
ajvb: Should this category be `httprequest` as well? With the subcategory, seems like it will be easy to filter these out when investigating something, but curious if it shouldn't have its own alert category.

ameihm0912: yeah I'm not sure, I was wondering about that too but this seemed the easiest -- the reason I did not make it it's own category is we can have multiple pipelines with the same resource indicator (e.g., two pipelines for one service) -- so this is a way to identify which pipeline the configuration is coming from.

ajvb: Yeah, I'm not clear on all of the implications of staying with this or making it unique, but maybe it could be pipeline specific? E.g. `httprequest-cfgtick` for the category? But that also seems potentially burdensome on the side of having to add each new one to anything reading cfgtick alerts.

ameihm0912: yeah this seems less impactful up front, I'll make a small adjustment and try that for now

PR #200, opened 2019-07-26
Add ALL available GuardDuty finding data to metadata

adrianosela:
This change should complete the aws guardduty side of the gatekeeper project

### DIFF
--- a/src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
+++ b/src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
@@ -1,6 +1,21 @@
 package com.mozilla.secops.gatekeeper;
 
+import com.amazonaws.services.guardduty.model.Action;
+import com.amazonaws.services.guardduty.model.AwsApiCallAction;
+import com.amazonaws.services.guardduty.model.City;
+import com.amazonaws.services.guardduty.model.Country;
+import com.amazonaws.services.guardduty.model.DnsRequestAction;
+import com.amazonaws.services.guardduty.model.DomainDetails;
 import com.amazonaws.services.guardduty.model.Finding;
+import com.amazonaws.services.guardduty.model.GeoLocation;
+import com.amazonaws.services.guardduty.model.LocalPortDetails;
+import com.amazonaws.services.guardduty.model.NetworkConnectionAction;
+import com.amazonaws.services.guardduty.model.Organization;
+import com.amazonaws.services.guardduty.model.PortProbeAction;
+import com.amazonaws.services.guardduty.model.PortProbeDetail;
+import com.amazonaws.services.guardduty.model.RemoteIpDetails;
+import com.amazonaws.services.guardduty.model.RemotePortDetails;
+import com.amazonaws.services.guardduty.model.Service;
 import com.mozilla.secops.IOOptions;
 import com.mozilla.secops.alert.Alert;
 import com.mozilla.secops.alert.AlertSuppressor;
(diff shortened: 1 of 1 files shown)


src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -112,9 +112,99 @@ public GenerateAlerts(Options opts) {
...
+          if (p.matcher(f.getType()).matches()) {
+            a.addMetadata("notify_email_direct", critNotifyEmail);
+            return;
+          }
+        }
+      }
+    }
+
+    private void addLocalPortMetadata(Alert a, LocalPortDetails lpd) {
+      a.addMetadata("local_port", Integer.toString(lpd.getPort()));
```
ameihm0912: Are these subfields always guaranteed to be non-null for example of `lpd` is non-null?

adrianosela: I was wondering if it would suffice (and be a good idea) to handle (not) setting a null value within the addMetadata() method?

ameihm0912: Yeah we should probably have it throw an exception or something, but I don't think we would want to silently ignore that, we would still want to check these values if they are important enough to add.

adrianosela:
What do you think about this approach? @ameihm0912 

Repeatedly using: 
```
    private void tryAddMeta(Alert a, String k, String v) {
      if (a == null || k == null || v == null) {
        return;
      }
      a.addMetadata(k, v);
    }
```

* Should this func be added to the Alert class as a method? 
* Should this func have arguments to optionally handle a null value?
* OR Should we instead modify the current addMetadata() method to throw an Exception?

ameihm0912:
>     * Should this func be added to the Alert class as a method?

Maybe at some point but I think we can leave that out for now?

>     * Should this func have arguments to optionally handle a null value?

What would this entail? Would that be like using a default value if value as `null` or something?

>     * OR Should we instead modify the current addMetadata() method to throw an Exception?

Yeah I think we should modify `addMetadata()` so it throws `IllegalArgumentException` or something on bad parameters but can do that one in another PR, it might take a bunch of changes to everything to catch the error.

adrianosela:
> > ```
> > * Should this func have arguments to optionally handle a null value?
> > ```
> 
> What would this entail? Would that be like using a default value if value was `null` or something?
> 

Im thinking of having a function as the last argument, and invoking it if the ```value``` argument is null and the function argument is not null

```
    private void tryAddMeta(Alert a, String k, String v, function nullValueHandler) {
      if (a == null || k == null || v == null) {
        if (nullValueHandler != null) {
            nullValueHandler()
        }
        return;
      }
      a.addMetadata(k, v);
    }
```

adrianosela:
Do not merge -- waiting on https://github.com/mozilla-services/foxsec-pipeline/pull/203 for a.tryAddMetadata() to exist in the Alert class

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
ajvb: As I commented here - https://github.com/mozilla-services/foxsec-pipeline/pull/203#issuecomment-516548528 - if `lpd.getPort()` returns `null`, `Integer.toString(lpd.getPort())` will throw a null pointer exception.

adrianosela:
This one's cleaned up ✅

PR #202, opened 2019-07-29
add transform docs for HTTPRequest to cfgtick

ameihm0912:


src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
```diff
@@ -476,6 +492,11 @@ public UserAgentBlacklistAnalysis(
       this.natView = natView;
     }
 
+    public String getTransformDoc() {
+      return new String(
+          "Alert if client makes request with user agent that matches entry in blacklist.");
```
adrianosela: nit: ```s/user agent that matches entry in blacklist/blacklisted user agent/``` for conciseness

PR #203, opened 2019-07-30
Implement tryAddMetadata()

adrianosela:
Currently, we call ```a.addMetadata(k, v)``` everywhere. In some cases, it is possible that v is null here. 

With the additions in this PR, it is no longer necessary to check whether a value is null before setting it, and we have more control over what to do if the value is in fact null

### Before:
```
String x = someObj.getSomeFieldWhichMayBeNull();
if (x != null) {
    a.addMetadata("the_x", x);
} else { // optionally if you need "the_x" to be set
    a.addMetadata("the_x", "some_default_unknown_value");
}
```
### Now:

If you know for a fact that x will not be null OR you dont care whether x gets set:

```
a.tryAddMetadata("the_x", someObj.getSomeFieldWhichMayBeNull());
```

Otherwise if you dont know if x will be null and you DO care that it gets set
```
try {
    a.mustAddMetadata("the_x", someObj.getSomeFieldWhichMayBeNull());
} catch (IllegalArgumentException e) {
    a.tryAddMetadata("the_x", "unknown");
}
```

### Why is this necessary?

In some cases when we want to add a lot of "interesting yet not critical" metadata to alerts, we find ourselves doing the following:
```
String asn = org.getAsn();
if (asn != null) {
    a.addMetadata("asn", asn);
}
String asnorg = org.getAsnOrg();
if (asnorg null) {
    a.addMetadata("asn_org", asnorg);
}
String isp = org.getIsp();
if (isp != null) {
    a.addMetadata("isp", isp);
}
String org = org.getOrg();
if (org != null) {
   a.addMetadata("org", org);
}
```

With this PR, the block above would become:

```
 a.tryAddMetadata("asn", org.getAsn());
 a.tryAddMetadata("asn_org", org.getAsnOrg());
 a.tryAddMetadata("isp", org.getIsp());
 a.tryAddMetadata("org", org.getOrg());
```

### DIFF
--- a/src/main/java/com/mozilla/secops/alert/Alert.java
+++ b/src/main/java/com/mozilla/secops/alert/Alert.java
@@ -248,6 +248,24 @@ public class Alert implements Serializable {
   }
 
   /**
+   * A best effort addMetadata(k,v)
+   *
+   * <p>If a null or empty key or value is provided, this function will not add anything. Do NOT use
+   * this function for metadata entries that are not purely informational.
+   *
+   * @param k String key
+   * @param v String value
+   * @return true if metadata entry was set, false otherwise
+   */
+  public boolean tryAddMetadata(String k, String v) {
+    if (k == null || v == null || k.equals("") || v.equals("")) {
+      return false;
+    }
+    addMetadata(k, v);
+    return true;
+  }
+
+  /**
    * Override alert timestamp
    *
    * @param timestamp Alert timestamp


ajvb:
To reduce the changes required, I would consider going about this the opposite way. Keep `addMetadata` as is and add something like `mustAddMetadata` that will throw an exception on null values.

As well though, relating to lines like 
```
tryAddMeta(a, "local_port", Integer.toString(lpd.getPort()));
```

this isn't actually a complete solution. If `lpg.getPort()` returns null, then `Integer.toString()` will throw a NullPointerException (https://stackoverflow.com/questions/27465731/string-valueof-vs-object-tostring/27465774).

adrianosela:
true, let me implement what you suggested, as well as figure out how to handle the NPE here -- i have an idea

adrianosela:
What do you think about it now? @ajvb This implementation also deals with the NullPointerException you pointed out:

```
a.tryAddMeta("local_port", Integer.toString(lpd.getPort()));
```

Becomes

```
a.tryAddMeta("local_port", lpd.getPort());
```

ajvb:
@adrianosela I think doing the `toString()` call within the `addMetadata` function makes sense. I don't think it makes sense to add a new function called `tryAddMeta` that already does what `addMetadata` does.

adrianosela:
just for the benefit of taking ```Object``` instead of ```String``` as the value arg @ajvb and deal with the possible NPE

src/main/java/com/mozilla/secops/alert/Alert.java
```diff
@@ -247,6 +247,48 @@ public void addMetadata(String key, String value) {
     }
   }
 
+  /**
+   * Add metadata entry or throw an exception on failure to do so
+   *
+   * @param k Key
+   * @param v Value
+   */
+  public void mustAddMetadata(String k, Object v) throws IllegalArgumentException {
```
ajvb: I don't think `mustAddMetadata` or `addMetadata` should accept `Object`. Let's make use of type safety here. You can make multiple funcs with the same name with different func signatures. So `mustAddMetadata(String k, String v)`, `mustAddMetadata(String k, Integer v)`, etc

ajvb: On the other side, if we want to accept `Object`, than we can make use of https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#toString()

You should be able to just do `addMetadata(k, v.toString())` and catch the NPE if `v` is `null`

ameihm0912: Yeah I'd agree on this one, let's get rid of the `Object` parameter here -- the existing function signature expects two `String` parameters (as do all the pipelines) so no reason to change this here right now for this.

src/main/java/com/mozilla/secops/alert/Alert.java
```diff
@@ -247,6 +247,36 @@ public void addMetadata(String key, String value) {
...
+    if (k == null) {
```
ameihm0912: In addition for checkout for `null` with both parameters should we also verify it is not an empty string?

src/main/java/com/mozilla/secops/alert/Alert.java
```diff
@@ -247,6 +247,36 @@ public void addMetadata(String key, String value) {
...
+    addMetadata(k, v.toString());
+  }
+
+  /**
+   * Add metadata entry or silently discard failure to do so
+   *
+   * @param k Key
+   * @param v Value
+   */
+  public void tryAddMetadata(String k, Object v) {
```
ameihm0912: I am thinking we just take this function out and leave it up to the pipeline to consume the errors if it wants to?

ajvb: agreed, this essentially a duplicate of `addMetadata`

adrianosela: The majority of the motivation behind this PR is not having to do a null check or catch an exception when setting a metadata value, since in most cases we repeatedly call addMetadata() as follows:

```
String asn = org.getAsn();
if (asn != null) {
    a.addMetadata("asn", asn);
}
String asnorg = org.getAsnOrg();
if (asnorg null) {
    a.addMetadata("asn_org", asnorg);
}
String isp = org.getIsp();
if (isp != null) {
    a.addMetadata("isp", isp);
}
String org = org.getOrg();
if (org != null) {
   a.addMetadata("org", org);
}
```

the code can get unnecessarily verbose

I wanted a change that allows us to call things like this --in one line-- (ignore the func name) with the guarantee that we wont end up with ```null``` as the value for any metadata field.

```
 a.addMetadata("asn", org.getAsn());
 a.addMetadata("asn_org", org.getAsnOrg());
 a.addMetadata("isp", org.getIsp());
 a.addMetadata("org", org.getOrg());
```

If we do what you both have suggested (keeping ```addMetadata(k,v)``` as is, and adding ```mustAddMetdata(k,v)```) we end up having to either:

- call addMetadata(k,v) after checking v is not null
- call addMetadata(k,v) without the guarantee of not setting a null value
- calling mustAddMetadata(k,v,) within a try/catch.

These three are just as verbose as the original usage of addMetadata(k,v)

@ajvb @ameihm0912 

Thoughts?

ameihm0912: @adrianosela ok yeah, so a best effort type of add

what if we were to adjust the function to return some sort of indication if it was added at least (maybe just a boolean return value, up to the caller to ignore), and adjust the function docs to indicate it's not intended for metadata that is not purely informational?

adrianosela: that sounds like a great idea 😄

adrianosela: Just to be clear, you are proposing we have ```mustAddMetadata(k,v)``` return a boolean, AND not throw an exception? @ameihm0912

ameihm0912: @adrianosela sorry am confused here now...

I thought what we were looking for here was a function to attempt to add metadata, but if the key or value or whatever is null or empty it would just return without doing anything.

What I was proposing prior was we implement a try function like you had suggested, but make it at least return `Boolean` (`true` if added, `false` if not). This would at least provide some clarity in the API documentation that it's possible the function will not add anything. And then we could also document that in the function javadoc to make sure it is obvious how it works.

or am I misunderstanding what we want here?

adrianosela: >  sorry am confused here now...
> 
> I thought what we were looking for here was a function to attempt to add metadata, but if the key or value or whatever is null or empty it would just return without doing anything.
> 

Yes 👍 

> What I was proposing prior was we implement a try function like you had suggested, but make it at least return `Boolean` (`true` if added, `false` if not). This would at least provide some clarity in the API documentation that it's possible the function will not add anything. And then we could also document that in the function javadoc to make sure it is obvious how it works.

Should I then get rid of the ```mustAddMetadata(k,v)``` currently implemented -- or do we still want that? 

@ameihm0912

ameihm0912: > Should I then get rid of the `mustAddMetadata(k,v)` currently implemented -- or do we still want that?

yeah if you don't mind, then this would unblock your other PR I believe

src/main/java/com/mozilla/secops/alert/Alert.java
```diff
@@ -247,6 +247,23 @@ public void addMetadata(String key, String value) {
...
   }
 
+  /**
+   * A best effort addMetadata(k,v)
+   *
+   * <p>If a null or empty key or value is provided, this function will not add anything. Do NOT use
+   * this function for metadata entries that are not purely informational.
+   *
+   * @param k Key
+   * @param v Value
```
ajvb: Missing `@return` java doc comment

PR #205, opened 2019-07-30
add release doc

ameihm0912:


docs/RELEASE.md
```diff
@@ -0,0 +1,81 @@
...
+Minor releases are used to release new features or more generally occur as part of the
+standard release cycle. A new minor release increments the minor release version number, and
+reverts the patch level number to `1`. As an example, the next minor release from `0.1.1` is
+`0.2.1`, and from `0.3.5` is `0.4.1`.
+
+New minor releases are created from `master`.
+
+### Regenerate documentation
+
+Regenerate javadoc documentation and merge the new documentation into `master`.
```
adrianosela: how?

ameihm0912: good point 👌 I'll add that before we merge this

ajvb:
@jvehent I've been thinking about that this week. I think it would be best to have constraints on each services API version(s) rather than the release's of the services/tools. And have these constraints/dependencies documented, verified, and tested. This way, you are focusing on how the different services communicate rather than how they may work internally.

jvehent:
If you roll back a disfunctioning pipeline, would you roll back all components to a known good version, or only individual components? If the former, how do you tie those working components together?

Issue #194, opened 2019-07-24
Include GCP Project Name and AWS Account Name in Gatekeeper alerts


ajvb:
It would be nice to include the AWS account name / the GCP project name in the Gatekeeper alerts for easier triaging.

For AWS, the IdentityManager can be used to translate the account id to account name:
https://github.com/mozilla-services/foxsec-pipeline/blob/3f68f65f517211d707cbbf13b619c1ece230a7ec/src/main/java/com/mozilla/secops/parser/Cloudtrail.java#L80-L92

For GCP, [`projects.get`](https://cloud.google.com/resource-manager/reference/rest/v1/projects/get) (or similar) can be used during alert creation to grab the project's metadata.

PR #209, opened 2019-08-01
Set GCP Project ID and AWS Account Name in metadata

adrianosela:
closes https://github.com/mozilla-services/foxsec-pipeline/issues/194

### DIFF
--- a/src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
+++ b/src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
@@ -4,17 +4,27 @@ import com.amazonaws.services.guardduty.model.Finding;
 import com.mozilla.secops.IOOptions;
 import com.mozilla.secops.alert.Alert;
 import com.mozilla.secops.alert.AlertSuppressor;
-import com.mozilla.secops.parser.*;
+import com.mozilla.secops.identity.IdentityManager;
+import com.mozilla.secops.parser.Event;
+import com.mozilla.secops.parser.GuardDuty;
+import com.mozilla.secops.parser.Payload;
+import java.io.IOException;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.regex.Pattern;
 import org.apache.beam.sdk.options.Default;
 import org.apache.beam.sdk.options.Description;
 import org.apache.beam.sdk.options.PipelineOptions;
-import org.apache.beam.sdk.transforms.*;
-import org.apache.beam.sdk.values.*;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.transforms.PTransform;
+import org.apache.beam.sdk.transforms.ParDo;
+import org.apache.beam.sdk.values.KV;
+import org.apache.beam.sdk.values.PCollection;
 import org.joda.time.DateTime;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** Implements various transforms on AWS GuardDuty {@link Finding} Events */
 public class GuardDutyTransforms implements Serializable {
@@ -101,7 +111,8 @@ public class GuardDutyTransforms implements Serializable {
     private static final String alertCategory = "gatekeeper:aws";
 
     private List<Pattern> escalate;
-    private String critNotifyEmail;
+    private final String critNotifyEmail;
+    private final String identityMgrPath;
 
     /**
      * static initializer for alert generation / escalation
@@ -110,8 +121,9 @@ public class GuardDutyTransforms implements Serializable {
      */
     public GenerateAlerts(Options opts) {
       critNotifyEmail = opts.getCriticalNotificationEmail();
-      String[] escalateRegexes = opts.getEscalateGDFindingTypeRegex();
+      identityMgrPath = opts.getIdentityManagerPath();
 
+      String[] escalateRegexes = opts.getEscalateGDFindingTypeRegex();
       escalate = new ArrayList<Pattern>();
       if (escalateRegexes != null) {
         for (String s : escalateRegexes) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -225,4 +230,72 @@ public void processElement(ProcessContext c) {
           .apply(ParDo.of(new AlertSuppressor(alertSuppressionWindow)));
     }
   }
+
+  public static class EnrichAlerts extends PTransform<PCollection<Alert>, PCollection<Alert>> {
```
ajvb: This looks pretty good, but I feel like it should just live in `GenerateAlerts`. We are always going to want this in live deployments of gatekeeper.

src/test/resources/testdata/gatekeeper/identitymanager.json
```diff
@@ -0,0 +1,5 @@
+{
```
ajvb: Lets just use https://github.com/mozilla-services/foxsec-pipeline/blob/master/src/test/resources/testdata/identitymanager.json

adrianosela: Will do, but I'd argue it is more modular this way. Otherwise very distinct tests depend on a single json file, adding/removing things for one test could be problematic and break others

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
adrianosela: ✅

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -4,17 +4,22 @@
 import com.mozilla.secops.IOOptions;
 import com.mozilla.secops.alert.Alert;
 import com.mozilla.secops.alert.AlertSuppressor;
+import com.mozilla.secops.identity.IdentityManager;
 import com.mozilla.secops.parser.*;
```
ameihm0912: Can we remove the wildcard import here?

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -120,12 +129,70 @@ public GenerateAlerts(Options opts) {
       } else {
         escalate.add(Pattern.compile(".+"));
       }
+
+      String idmgrPath = opts.getIdentityManagerPath();
+      if (idmgrPath != null) {
+        try {
+          IdentityManager idmgr = IdentityManager.load(idmgrPath);
+          if (idmgr != null) {
```
ameihm0912: I don't believe this is possible, you will just end up with an `IOException` if there is an issue

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -120,12 +129,70 @@ public GenerateAlerts(Options opts) {
...
+            awsAcctMap = idmgr.getAwsAccountMap();
+            if (awsAcctMap != null) {
+              log.info("aws account map successfully loaded from identity manager file");
+            } else {
+              log.warn(
+                  "no aws account map contained in identity manager file, alerts will have \"UNKNOWN\" as aws_account_name");
+            }
+          } else {
+            log.warn("bad identity manager file, alerts will have \"UNKNOWN\" as aws_account_name");
```
ameihm0912: This should likely be `error`

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -120,12 +129,70 @@ public GenerateAlerts(Options opts) {
...
+          }
+        } catch (IOException x) {
+          log.warn(
```
ameihm0912: Similar here

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -120,12 +129,70 @@ public GenerateAlerts(Options opts) {
       } else {
         escalate.add(Pattern.compile(".+"));
       }
+
+      String idmgrPath = opts.getIdentityManagerPath();
```
ameihm0912: I'm wondering if this should be done here, or if instead it should be done in `setup` for the `ParDo` later on; see also https://github.com/mozilla-services/foxsec-pipeline/blob/62d7d77b50ccf2016f7b6fe601ddcf18feb5591b/src/main/java/com/mozilla/secops/authprofile/AuthProfile.java#L360-L364

If we do it as part of `DoFn` setup this makes it much easier to change to be refreshed periodically/etc later

adrianosela: 👍

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -100,8 +110,11 @@ public void processElement(ProcessContext c) {
 
     private static final String alertCategory = "gatekeeper:aws";
 
+    private Map<String, String> awsAcctMap;
```
ameihm0912: This could be moved down to live in the anonymous `DoFn` class you define later since it's not actually used in `GuardDutyTransforms` itself

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -100,8 +110,11 @@ public void processElement(ProcessContext c) {
...
+
     private List<Pattern> escalate;
     private String critNotifyEmail;
+    private String identityMgrPath;
```
ameihm0912: These can probably be marked `final`

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -122,10 +136,44 @@ public GenerateAlerts(Options opts) {
...
-        a.addMetadata("notify_email_direct", critNotifyEmail);
+        for (Pattern p : escalate) {
+          if (p.matcher(findingType).matches()) {
+            a.addMetadata("notify_email_direct", critNotifyEmail);
+          }
+        }
+      }
+    }
+
+    private void tryAddAccountName(Alert a) {
```
ameihm0912: Should this be moved into the anonymous `DoFn` too?

adrianosela:
Comments addressed and commits squashed

PR #210, opened 2019-08-01
Add cfgticks to AuthProfile

ajvb:
Includes cfgticks for both StateAnalyze and CritObjectAnalyze.

@ameihm0912 Also, should I rename the `DocumentingTransform` interface since in this case it's being used on DoFn's?

### DIFF
--- a/src/main/java/com/mozilla/secops/authprofile/AuthProfile.java
+++ b/src/main/java/com/mozilla/secops/authprofile/AuthProfile.java
@@ -2,6 +2,7 @@ package com.mozilla.secops.authprofile;
 
 import com.mozilla.secops.CidrUtil;
 import com.mozilla.secops.CompositeInput;
+import com.mozilla.secops.DocumentingTransform;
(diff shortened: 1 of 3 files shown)


adrianosela:
I think we should rename ```DocumentingTransform``` to  just ```Documented```

src/main/java/com/mozilla/secops/authprofile/AuthProfile.java
```diff
@@ -241,6 +255,12 @@ public CritObjectAnalyze(AuthProfileOptions options) {
       critNotifyEmail = options.getCriticalNotificationEmail();
     }
 
+    public String getTransformDoc() {
+      return String.format(
+          "Alert via %s immediately on auth events to specified objects: %s",
```
adrianosela: I'd say something like 

```
return String.format("notify email %s on authentication events against any of: %s", critNotifyEmail, Arrays.toString(critObjects));
```

PR #212, opened 2019-08-02
Parsing Auth0 events from auth0pull

ajvb:
Adds an Auth0 parser that makes use of the official auth0 mgmt sdk.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -127,7 +127,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.datatype</groupId>
             <artifactId>jackson-datatype-joda</artifactId>
-	    <version>2.9.9</version>
+            <version>2.9.9</version>
         </dependency>
         <dependency>
             <groupId>org.apache.httpcomponents</groupId>
@@ -187,7 +187,7 @@
         <dependency>
             <groupId>net.spy</groupId>
             <artifactId>spymemcached</artifactId>
-	    <version>2.12.3</version>
+            <version>2.12.3</version>
         </dependency>
         <dependency>
             <groupId>commons-cli</groupId>
@@ -249,6 +249,11 @@
             <artifactId>groovy-all</artifactId>
             <version>2.5.6</version>
             <type>pom</type>
+        </dependency>
+        <dependency>
+            <groupId>com.auth0</groupId>
+            <artifactId>auth0</artifactId>
+            <version>1.14.1</version>
         </dependency>
     </dependencies>
 
--- a/src/main/java/com/mozilla/secops/parser/Auth0.java
+++ b/src/main/java/com/mozilla/secops/parser/Auth0.java
@@ -1,0 +1,246 @@
+package com.mozilla.secops.parser;
+
+import com.auth0.json.mgmt.logevents.LogEvent;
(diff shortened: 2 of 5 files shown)


src/main/java/com/mozilla/secops/parser/Payload.java
```diff
@@ -8,6 +8,8 @@
 
   /** Type of payload data stored */
   public enum PayloadType {
+    /** Auth0 log event */
+    AUTH0,
```
ameihm0912: Is this going to cause problems with a pipeline update that contains in-flight event information? I've been trying to add new event types towards the end which bumps raw down but keeps the values associated with the previous types

src/main/java/com/mozilla/secops/parser/Auth0.java
```diff
@@ -0,0 +1,239 @@
...
+    AuthTypes.add("s");
+    AuthTypes.add("ssa");
+    AuthTypes.add("seacft");
+    AuthTypes.add("scoa");
+  }
+
+  private static Map<String, String> TypeMap;
+
+  // Mapping of auth0 type codes to there name
+  // https://auth0.com/docs/logs#log-data-event-listing
```
adrianosela: Would it be worth including all codes so that  ```isAuthEvent()``` does not return false on an unlisted code?
Also, question regarding missing codes here e.g. seoobft and seotpft... why have you chosen to leave them out - just dont expect to ever get any of those?

src/main/java/com/mozilla/secops/parser/Auth0.java
```diff
@@ -0,0 +1,239 @@
...
+   *
+   * @return Username as string, or null if it can't be found.
+   */
+  public String getUsername() {
+    if (event.getDetails() != null && event.getDetails().containsKey("prompts")) {
+      if (event.getDetails().get("prompts") instanceof ArrayList) {
+        // The compiler cannot check that the parameter of `Object` is actually safe,
+        // therefore it will always throw a warning here (and below) even though they are
+        // done in a safe manner.
+        @SuppressWarnings("unchecked")
```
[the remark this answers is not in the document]
ajvb: Ah cool, yeah absolutely.

pom.xml
```diff
@@ -127,7 +127,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.datatype</groupId>
             <artifactId>jackson-datatype-joda</artifactId>
-	    <version>2.9.9</version>
+            <version>2.9.9</version>
```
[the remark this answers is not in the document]
ajvb: yep

src/main/java/com/mozilla/secops/parser/Payload.java
ajvb: Ah good point. Fixing.

src/main/java/com/mozilla/secops/parser/Auth0.java
ajvb: Just missed those codes. Thanks, adding them now.

adrianosela: there may be more missing ones than the two I mentioned

PR #213, opened 2019-08-02
remove default value for aws_account_name metadata entry

adrianosela:
Currently, when an identitymanager.json file is not provided, we set the aws_account_name metadata entry to UNKNOWN. I thought this would be useful but in reality its just polluting alerts for no good reason.

This PR removes setting this default value, and refactors where the logger of the GenerateAlerts class is initialized

### DIFF
--- a/src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
+++ b/src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
@@ -129,12 +129,16 @@ public class GuardDutyTransforms implements Serializable {
     private final String critNotifyEmail;
     private final String identityMgrPath;
 
+    private Logger log;
+
     /**
      * static initializer for alert generation / escalation
      *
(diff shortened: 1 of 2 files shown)


src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -334,30 +338,27 @@ private void tryAddAccountName(Alert a) {
...
-
                   if (identityMgrPath != null) {
                     try {
                       awsAcctMap = IdentityManager.load(identityMgrPath).getAwsAccountMap();
                       if (awsAcctMap != null) {
                         log.info("aws account map successfully loaded from identity manager file");
                       } else {
                         log.warn(
-                            "no aws account map contained in identity manager file, alerts will have \"UNKNOWN\" as aws_account_name");
+                            "no aws account map contained in identity manager file, alerts will not contain have aws_account_name");
```
ameihm0912: typo here

Issue #182, opened 2019-07-18
Develop standardized method for packing an array of values into a metadata key


ameihm0912:
See also comment in #181 

Some elements of the pipeline put a list of values under a metadata key, right now it is essentially free form strings and up to the producer to properly delimit the elements.

That can be improved so there is a consistent method for putting a list of values into metadata.

PR #214, opened 2019-08-04
Add all available data in ETD finding to alert metadata + test all metadata

adrianosela:


src/main/java/com/mozilla/secops/gatekeeper/ETDTransforms.java
```diff
@@ -127,9 +131,71 @@ public GenerateAlerts(Options opts) {
...
+    }
+
+    private void tryAddEscalationEmail(Alert a, EventThreatDetectionFinding f) {
+
+      String ruleName;
+      try {
+        ruleName = f.getDetectionCategory().getRuleName();
+      } catch (NullPointerException e) {
+        return;
+      }
```
ajvb: This is totally fine, but to stay consistent with the rest of the code base it would be worth changing this to:
```
if (f.getDetectionCategory() == null || f.getDetectionCategory().getRuleName() == null) {
  return;
}
```

src/main/java/com/mozilla/secops/gatekeeper/ETDTransforms.java
```diff
@@ -127,9 +131,71 @@ public GenerateAlerts(Options opts) {
...
+        for (Pattern p : escalate) {
+          if (p.matcher(ruleName).matches()) {
+            a.addMetadata("notify_email_direct", critNotifyEmail);
+            return;
+          }
+        }
+      }
+    }
+
+    // add informational metadata using values within finding.
```
ajvb: Might be best to write this comment in java doc format?

src/main/java/com/mozilla/secops/gatekeeper/ETDTransforms.java
```diff
@@ -127,9 +131,71 @@ public GenerateAlerts(Options opts) {
...
+              "suspicious activity detected in gcp org %s project %s",
+              f.getSourceId().getCustomerOrganizationNumber(), f.getProperties().getProject_id()));
+      a.setTimestamp(DateTime.parse(f.getEventTime()));
+      a.setCategory(alertCategory);
+      a.setSeverity(Alert.AlertSeverity.CRITICAL);
+      if (f.getProperties().getLocation() != null) {
+        a.addMetadata("location", f.getProperties().getLocation());
+      }
+      a.addMetadata("indicator", f.getDetectionCategory().getIndicator());
+      a.addMetadata("rule_name", f.getDetectionCategory().getRuleName());
```
ajvb: Why the null check for `f.getDetectionCategory().getRuleName()` in `tryAddEscalationEmail()` but not here? Might want to add null checks in here

adrianosela: yeah, thats one of the fields which are guaranteed to be null. But I will add null checks everywhere possible. There are some similar ones missing in the guardduty stuff. Will update both

src/main/java/com/mozilla/secops/gatekeeper/ETDTransforms.java
```diff
@@ -127,9 +132,86 @@ public GenerateAlerts(Options opts) {
...
+      a.tryAddMetadata("detection_timestap", f.getEventTime());
+      Properties prop = f.getProperties();
+      if (prop != null) {
+        a.tryAddMetadata("subnetwork_id", prop.getSubnetwork_id());
+        a.tryAddMetadata("subnetwork_name", prop.getSubnetwork_name());
+        a.tryAddMetadata("ip", prop.getIp());
+        List<String> doms = prop.getDomain();
+        if (doms != null) {
+          for (String d : doms) {
+            a.tryAddMetadata("domain", d);
```
ajvb: Relating to #182 - Not sure about this. It can make certain SQL query results very strange, but also not sure about the current approach of packing it into a comma separated string a la https://github.com/mozilla-services/foxsec-pipeline/pull/181#discussion_r305110349

ajvb: @ameihm0912 What are your thoughts on this?

ameihm0912: Yeah I'm not too sure on this either, and it does sound like it will present problems with BQ. This will also cause problems if the metadata ever needs to be fetched from the `Alert` object in the pipeline, since we can't return an array of values there right now.

Like AJ mentioned, we handle it like this right now (which is also not ideal) but might as well be consistent until we can fix it across the board.

https://github.com/mozilla-services/foxsec-pipeline/blob/37604779ece1e3a26559e9bf00cead4e1c1d5f68/src/main/java/com/mozilla/secops/amo/AddonMatcher.java#L115

src/main/java/com/mozilla/secops/gatekeeper/ETDTransforms.java
```diff
@@ -127,9 +132,86 @@ public GenerateAlerts(Options opts) {
...
+      Properties prop = f.getProperties();
+      if (prop != null) {
+        a.tryAddMetadata("project_id", prop.getProject_id());
+        a.tryAddMetadata("location", prop.getLocation());
+      }
+      if (sid != null
+          && prop != null
+          && sid.getCustomerOrganizationNumber() != null
+          && prop.getProject_id() != null) {
+        a.setSummary(
```
ameihm0912: Does this mean the summary will be conditionally set? The summary should always be set to something.

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -154,19 +154,26 @@ public GenerateAlerts(Options opts) {
...
+      a.tryAddMetadata("aws_region", f.getRegion());
+      a.tryAddMetadata("description", f.getDescription());
+      Double severity = f.getSeverity();
+      if (severity != null) {
+        a.tryAddMetadata("finding_aws_severity", Double.toString(severity));
+      }
+      a.tryAddMetadata("finding_type", f.getType());
+      a.tryAddMetadata("finding_id", f.getId());
+      if (f.getAccountId() != null && f.getTitle() != null) {
+        a.setSummary(
```
ameihm0912: Similar issue as described prior here.

src/main/java/com/mozilla/secops/gatekeeper/ETDTransforms.java
ameihm0912: discussed with @adrianosela in more detail; these repeated fields have been tested in BQ and present challenges trying to group them in the same value in some cases (not with the `domain` element, but with others that are an array of maps essentially)

given that and we are already doing this in GD this is probably OK then, but another reason we will need to find a good solution for #182 soon so we can handle this better

PR #215, opened 2019-08-06
custom beam metric for iprepd violation writes

ameihm0912:


src/test/java/com/mozilla/secops/amo/TestAmo.java
```diff
@@ -155,6 +160,22 @@ public void amoFxaAbuseNewVersionTest() throws Exception {
...
+    pResult.waitUntilFinish();
+
+    Iterable<MetricResult<Long>> vWrites =
+        pResult
+            .metrics()
+            .queryMetrics(
+                MetricsFilter.builder()
+                    .addNameFilter(
+                        MetricNameFilter.named(
+                            "com.mozilla.secops.IprepdIO$WriteFn",
```
ajvb: Did this show up in stackdriver? Looking at https://cloud.google.com/monitoring/api/v3/metrics-details#label_names - I worry that the `$` will not work (related: https://stackoverflow.com/a/48566003)

ameihm0912: @ajvb interesting, nice catch -- I did not test the metrics showing up in Stackdriver.

Most of the examples I have seen register the class name as the namespace, so I just went with that but perhaps that will not work properly with inner classes in Stackdriver.

I've modified this PR so `IprepdIO` defines it's own namespace, and also fixed a bug in the tests.

Issue #220, opened 2019-08-12
Add EC2 instance name (app tag) whitelist support to Gatekeeper Pipeline


adrianosela:
So far Guard Duty has been pretty good at classifying *ALL* network connection type (brute force attempts, port scans, etc) as low severity. 

However, it generates many high severity "DNS REQUEST" type findings for EC2 instances querying bad domains (crypto, phishing, malware, etc).

These are **NOT** false positives, but in most cases we don't care about them because they are expected due to the nature of the running host.

For example: we run a ton of web crawlers to asses the state of the web in general, etc, these hit those bad domains all the time, and we dont care.

We should have a way of suppressing these type of alerts for these EC2 instances. That should not be hard given the fact that the ec2 instance tags are contained in the finding object.

CC:// @ameihm0912  @ajvb

PR #223, opened 2019-08-15
[Depends on 222] Gatekeeper: Suppress DNS_REQUEST findings by EC2 instance id

adrianosela:
depends on https://github.com/mozilla-services/foxsec-pipeline/pull/222 and closes https://github.com/mozilla-services/foxsec-pipeline/issues/220.

The context on why this is needed is in https://github.com/mozilla-services/foxsec-pipeline/issues/220 -- have a look before you review the PR


adrianosela:
Closing -- we will implement the same DNS_REQUEST finding suppression, but by app tag, not by EC2 inst ID

PR #224, opened 2019-08-15
suppress dns findings for specific apps given app tag

adrianosela:
closes https://github.com/mozilla-services/foxsec-pipeline/issues/220


src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -116,6 +146,14 @@ public void processElement(ProcessContext c) {
                       return;
                     }
                   }
+                  String appTag = extractAppTag(f);
+                  if (ignoreDNSFindingAppNames != null && appTag != null) {
+                    for (String whitelisted : ignoreDNSFindingAppNames) {
+                      if (whitelisted.equals(appTag)) {
+                        return;
```
ameihm0912: Is this intended to apply only to `DNS_REQUEST` findings? If so, does it? I can't tell from the diff but it looks like it applies to everything.

adrianosela: good catch

Issue #207, opened 2019-07-31
Persist Guardduty alerts to either Bigquery or GCS


ajvb:
As we discussed, it would be helpful to persist all Guardduty alerts in there raw json to either a separate BigQuery table or to GCS.

The goal is to be able to dig into alerts easily during an investigation.

adrianosela:
The "payload" field in my alerts is still empty (```null```), would this be a good place to throw the raw events? @ajvb 

cc:// @ameihm0912

ajvb:
@adrianosela No, I think that it should be persisted outside of the context of foxsec-pipeline alerts. To be more explicit, we should persist the `com.amazonaws.services.guardduty.model.Finding` parsed within


https://github.com/mozilla-services/foxsec-pipeline/blob/49e28bf903515dc48ebe7767709ee947c9710927/src/main/java/com/mozilla/secops/parser/GuardDuty.java#L11-L22

to it's own BigQuery table or to GCS.

adrianosela:
Gotcha, will implement a "Persistor" transform for each of ETD and GD. Will take the target BQ as an argument i.e. ```--persistETDFindingsBigQuery``` and ```--persistGDFindingsBigQuery``` and store them separately.

Does this sound like a reasonable approach? @ajvb @ameihm0912 

Also, what should the behavior be if the user fails to provide these arguments? Run anyways and not persist, or die on deploy?

ajvb:
Let's just do Guardduty. For ETD, we can use [CSCC](https://cloud.google.com/security-command-center/).

I'd base it off of https://github.com/mozilla-services/foxsec-pipeline/blob/master/src/main/java/com/mozilla/secops/OutputOptions.java#L23-L26 and https://github.com/mozilla-services/foxsec-pipeline/blob/master/src/main/java/com/mozilla/secops/CompositeOutput.java#L88-L107

So you should probably just be able to use `org.apache.beam.sdk.io.gcp.bigquery.BigQueryIO` for this.

adrianosela:
Closing in favor of https://github.com/mozilla-services/foxsec-pipeline/issues/229 @ajvb

PR #225, opened 2019-08-16
gatekeeper: persist guardduty findings in big query

adrianosela:
closes https://github.com/mozilla-services/foxsec-pipeline/issues/207,

Updates in terraform: 

- NEW big query dataset for guardduty findings
- NEW big query table for guardduty findings, with schema:
```
[
  {
    "name": "awsAccountId",
    "type": "STRING",
    "mode": "NULLABLE"
  },
  {
    "name": "guarddutyFindingId",
    "type": "STRING",
    "mode": "NULLABLE"
  },
  {
    "name": "guarddutyFindingType",
    "type": "STRING",
    "mode": "NULLABLE"
  },
  {
    "name": "raw",
    "type": "STRING",
    "mode": "NULLABLE"
  },
  {
    "name": "timestamp",
    "type": "TIMESTAMP",
    "mode": "NULLABLE"
  }
]
```

Updates in deployment script:

- NEW argument: 
```--findingsPersistenceBigQuery=${GOOGLE_PROJECT_ID}:${PERSIST_DATASET_NAME}.${PERSIST_TABLE_NAME}``` 

Both these things live in https://github.com/mozilla-services/secops-gatekeeper-poc master branch

### NOTES: this is not yet deployed in dataflow, but the terraform has been ran

### DIFF
--- a/src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
+++ b/src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
@@ -20,6 +20,9 @@ import com.amazonaws.services.guardduty.model.RemotePortDetails;
 import com.amazonaws.services.guardduty.model.Resource;
 import com.amazonaws.services.guardduty.model.Service;
 import com.amazonaws.services.guardduty.model.Tag;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.api.services.bigquery.model.TableRow;
 import com.mozilla.secops.IOOptions;
 import com.mozilla.secops.alert.Alert;
 import com.mozilla.secops.alert.AlertSuppressor;
@@ -33,6 +36,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
+import org.apache.beam.sdk.io.gcp.bigquery.BigQueryIO;
 import org.apache.beam.sdk.options.Default;
 import org.apache.beam.sdk.options.Description;
 import org.apache.beam.sdk.options.PipelineOptions;
@@ -41,6 +45,7 @@ import org.apache.beam.sdk.transforms.PTransform;
 import org.apache.beam.sdk.transforms.ParDo;
 import org.apache.beam.sdk.values.KV;
 import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PDone;
 import org.joda.time.DateTime;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java [hunk 1]
```diff
@@ -122,6 +132,78 @@ public void processElement(ProcessContext c) {
     }
   }
 
+  /** Persist GuardDuty findings onto GCP's Big Query */
+  public static class PersistFindings extends PTransform<PCollection<Event>, PDone> {
+    private static final long serialVersionUID = 1L;
+
+    private static ObjectMapper mapper = new ObjectMapper();
+    private static String findingsPersistenceBigQuery;
```
ameihm0912: Any reason this needs to be static?

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -122,6 +132,78 @@ public void processElement(ProcessContext c) {
...
+    private TableRow FindingsTableRow(Finding f) {
+      TableRow r = new TableRow();
+      r.set("awsAccountId", f.getAccountId());
+      r.set("guarddutyFindingId", f.getId());
+      r.set("guarddutyFindingType", f.getType());
+      r.set("timestamp", f.getUpdatedAt());
+
+      String raw;
+      try {
+        raw = mapper.writeValueAsString(f);
```
ameihm0912: Is this the only way to do this? Is some sort of conversion from the JSON to a row that would create a column for each field not possible?

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -122,6 +132,78 @@ public void processElement(ProcessContext c) {
...
+     *
+     * @param opts {@link Options} pipeline options
+     */
+    public PersistFindings(Options opts) {
+      findingsPersistenceBigQuery = opts.getFindingsPersistenceBigQuery();
+    }
+
+    @Override
+    public PDone expand(PCollection<Event> input) {
+      if (findingsPersistenceBigQuery != null) {
```
ameihm0912: Can we just remove this `if`? This transform should only be initialized when this option is set anyway no?

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
adrianosela: nope 👍

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
adrianosela: yes we can 👍

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
ajvb: I'll just call out that this was per my suggestion. My thinking here was that dealing with migrations if/when new GD fields were changed/removed/added might be annoying and we initially just wanted a centralized place to fetch the raw GD finding.

ameihm0912: Ok yeah, I wasn't sure if BQ had a nice way to handle arbitrary JSON or something like this, but that makes sense

adrianosela:
> I feel as though this would be nicer if it was implemented at the global level, possibly by extending event payload classes to support something like `toTableRow`, and then just implementing this functionality in `Composite{Input,Output}`.

opened https://github.com/mozilla-services/foxsec-pipeline/issues/229 and closing this PR @ajvb @ameihm0912

Issue #229, opened 2019-08-19
Persist ALL events somewhere (e.g. BigQuery)


adrianosela:
We aren't persisting anything outside of the built-in persistence options of the data sources we read from.

Fulfilling this issue involves storing all generic events somewhere

ameihm0912:
@adrianosela this is an architectural change that will be fairly involved; I am closing this issue for now until we have had a chance to discuss the approach and we will recreate it with the requirements for this once we have them

adrianosela:
Gotcha. Just wanted to make sure I was persisting GuardDuty events somehow before my departure, however if you're planning on implementing this in a way that doesnt require changes to individual pipelines I guess it is no longer my duty to get this done before then. @ameihm0912

PR #230, opened 2019-08-19
composite persistor with bigquery as initial persistor

adrianosela:
closes https://github.com/mozilla-services/foxsec-pipeline/issues/229

This PR:

- **New** PersistenceOptions interface to get external datastore options where to persist events
- **New** CompositePersitence class which persists events in all provided persistors
- **New** PersistInBigQuery class (transform) which persists events in BigQuery
- Using composite persistence in the Gatekeeper Pipeline

### DIFF
--- a/src/main/java/com/mozilla/secops/PersistenceOptions.java
+++ b/src/main/java/com/mozilla/secops/PersistenceOptions.java
@@ -1,0 +1,11 @@
+package com.mozilla.secops;
+
+import org.apache.beam.sdk.options.Description;
+import org.apache.beam.sdk.options.PipelineOptions;
+
+public interface PersistenceOptions extends PipelineOptions {
+  @Description("Write payloads to BigQuery; BigQuery table specification")
+  String getPersistenceBigQuery();
+
+  void setPersistenceBigQuery(String value);
+}
--- a/src/main/java/com/mozilla/secops/gatekeeper/GatekeeperPipeline.java
+++ b/src/main/java/com/mozilla/secops/gatekeeper/GatekeeperPipeline.java
@@ -1,6 +1,7 @@
 package com.mozilla.secops.gatekeeper;
 
 import com.mozilla.secops.CompositeInput;
+import com.mozilla.secops.CompositePersistence;
 import com.mozilla.secops.OutputOptions;
 import com.mozilla.secops.alert.Alert;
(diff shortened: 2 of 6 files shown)


src/main/java/com/mozilla/secops/PersistenceOptions.java
```diff
@@ -0,0 +1,11 @@
+package com.mozilla.secops;
+
+import org.apache.beam.sdk.options.Description;
+import org.apache.beam.sdk.options.PipelineOptions;
+
+public interface PersistenceOptions extends PipelineOptions {
```
ajvb: I don't think this needs its own interface. Can it be placed within `OutputOptions`?

src/main/java/com/mozilla/secops/gatekeeper/GatekeeperPipeline.java
```diff
@@ -40,6 +41,8 @@
             .apply("parse input", new GatekeeperParser.Parse(options))
             .apply("window input", new GlobalTriggers<Event>(60));
 
+    events.apply("persist events", new CompositePersistence(options));
```
ameihm0912: The objective would be to avoid needing to do this in the pipelines if we make it a generic option, it should be handled transparently to the pipeline itself

src/main/java/com/mozilla/secops/persistence/PersistInBiqQuery.java
```diff
@@ -0,0 +1,65 @@
...
+  public PersistInBiqQuery(String bigQuerySpec) {
+    persistenceBQ = bigQuerySpec;
+  }
+
+  private TableRow buildTableRow(Event e) {
+    TableRow r = new TableRow();
+    r.set("event_id", e.getEventId());
+    r.set("event_type", e.getPayloadType().name());
+    r.set("event_time", e.getTimestamp());
+    r.set("raw", e.getPayload().toString());
```
ameihm0912: I suspect this likely will not work with most things, this will require larger changes across all the event types as IIRC most of them do not support serialization to JSON; we likely may also want to include the raw event if we do this and not just the parsed event

adrianosela:
Closing as per the discussion in https://github.com/mozilla-services/foxsec-pipeline/issues/229. Thanks for the reviews.

PR #233, opened 2019-08-23
IPrepd applied violation logs parser

adrianosela:
a basic parser for applied violation logs from IPrepd

### DIFF
--- a/src/main/java/com/mozilla/secops/parser/IPrepdLog.java
+++ b/src/main/java/com/mozilla/secops/parser/IPrepdLog.java
@@ -1,0 +1,142 @@
+package com.mozilla.secops.parser;
+
+import java.io.Serializable;
+import java.util.Map;
+
+/** Payload parser for IPrepd logs */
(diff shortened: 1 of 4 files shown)


src/main/java/com/mozilla/secops/parser/IPrepdLog.java
```diff
@@ -0,0 +1,142 @@
...
+   */
+  public IPrepdLog(String input, Event e, ParserState s) {
+    Map<String, Object> fields = Parser.convertJsonToMap(input);
+    if (fields == null) {
+      return;
+    }
+    msg = fields.get("msg").toString();
+    violation = fields.get("violation").toString();
+    decayAfter = fields.get("decay_after").toString();
+    originalReputation = Integer.parseInt(fields.get("original_reputation").toString());
```
adrianosela: hmmm... these should probably be casts -- will fix

adrianosela: ✅

PR #237, opened 2019-08-29
Input multiplex mode support

ameihm0912:
This is the bulk of the remaining changes for multi-input filtering.

This adds support for multiplex mode to `Input`, including both raw multi-element reads and parsed multi-element reads, where an element can have more than one input stream. Filters can be applied to incoming event streams on an element basis.


src/main/java/com/mozilla/secops/input/InputElement.java
```diff
@@ -63,48 +83,99 @@
     return list.apply(Flatten.<String>pCollections());
   }
 
+  public PCollection<Event> expandElement(PBegin begin) {
+    if (parserCfg == null) {
+      throw new RuntimeException("no parser configured for element");
```
ajvb: non-blocking: I/we haven't been doing this much/any in our tests, but it might be nice to start including failure tests (a la https://stackoverflow.com/a/6527990) for these exceptions.

ameihm0912: yeah agreed 👍

PR #240, opened 2019-09-06
Adds cfgticks for Gatekeeper + some clean up

ajvb:
Adds support for cfgticks to Gatekeeper. Also cleans some pieces up and adds support for disabling ETD or GD.

### DIFF
--- a/src/main/java/com/mozilla/secops/gatekeeper/ETDTransforms.java
+++ b/src/main/java/com/mozilla/secops/gatekeeper/ETDTransforms.java
@@ -1,5 +1,6 @@
 package com.mozilla.secops.gatekeeper;
 
+import com.mozilla.secops.DocumentingTransform;
 import com.mozilla.secops.IOOptions;
 import com.mozilla.secops.alert.Alert;
 import com.mozilla.secops.alert.AlertSuppressor;
@@ -105,7 +106,8 @@ public class ETDTransforms implements Serializable {
   }
 
   /** Generate Alerts for relevant ETD Finding Events */
-  public static class GenerateAlerts extends PTransform<PCollection<Event>, PCollection<Alert>> {
+  public static class GenerateETDAlerts extends PTransform<PCollection<Event>, PCollection<Alert>>
+      implements DocumentingTransform {
     private static final long serialVersionUID = 1L;
 
     private static final String alertCategory = "gatekeeper:gcp";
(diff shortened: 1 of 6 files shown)


src/main/java/com/mozilla/secops/input/InputElement.java
```diff
@@ -58,10 +58,10 @@ public String getName() {
     }
 
     for (String i : fileInputs) {
-      list = list.and(begin.apply(TextIO.read().from(i)));
+      list = list.and(begin.apply(i, TextIO.read().from(i)));
```
ajvb: @ameihm0912 I don't think this is the right way to fix this, but wanted to push it up to get tests passing and to discuss. With how this is currently setup, when using multiple text files with `Input.compositeInputAdapter` we get this exception:
```
Pipeline update will not be possible because the following transforms do not have stable unique names: input/TextIO.Read.

Conflicting instances:
- name=input/TextIO.Read:
    - TextIO.Read
    - TextIO.Read

You can fix it adding a name when you call apply(): pipeline.apply(<name>, <transform>).
        at com.mozilla.secops.gatekeeper.TestGatekeeper.gatekeeperNoFiltersTest(TestGatekeeper.java:214)
```

I imagine this will also blow up when we deploy any pipeline using `Input.compositeInputAdapter` as there is the same issue below with PubsubIO.

ameihm0912: @ajvb this should actually be fixed in the PR I am going to submit shortly, I ran into the same issue

src/main/java/com/mozilla/secops/gatekeeper/ETDTransforms.java
```diff
@@ -132,6 +134,10 @@ public GenerateAlerts(Options opts) {
       }
     }
 
+    public String getTransformDoc() {
+      return "Alerts are generated based on events sent from GCP's Event Threat Detection.";
```
ameihm0912: Should we include the matching regex? I guess that isn't really required if it's in pipeline options, but I think current metrics only look at heuristics right now

cc @psiinon

src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java
```diff
@@ -199,6 +201,10 @@ public GenerateAlerts(Options opts) {
       }
     }
 
+    public String getTransformDoc() {
+      return "Alerts are generated based on events sent from AWS's Guardduty.";
```
ameihm0912: Same comment as previous here

PR #242, opened 2019-09-09
Some cleanup and optimizations for multimode HTTPRequest

ameihm0912:
This is a follow up to #241 that contains a bunch of cleanup and optimizations.

This also has a fix to restore being able to run a specific method in a test rather than the entire test class.

### DIFF
--- a/src/main/java/com/mozilla/secops/DetectNat.java
+++ b/src/main/java/com/mozilla/secops/DetectNat.java
@@ -34,22 +34,6 @@ public class DetectNat extends PTransform<PCollection<Event>, PCollection<KV<Str
 
   private static final Long UAMARKPROBABLE = 2L;
 
-  private final String stepPrefix;
-
-  /**
-   * Create new DetectNat
-   *
-   * @param stepPrefix Prefix for transform step or null for none
-   */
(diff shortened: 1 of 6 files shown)


src/main/java/com/mozilla/secops/DetectNat.java
```diff
@@ -68,29 +52,18 @@ public DetectNat() {
...
-  public static PCollectionView<Map<String, Boolean>> getView(
-      PCollection<Event> events, String prefix) {
-    if (prefix == null) {
-      prefix = "";
-    }
-    return events
-        .apply(String.format("%s nat view", prefix), new DetectNat(prefix))
-        .apply(String.format("%s nat map view", prefix), View.<String, Boolean>asMap());
+  public static PCollectionView<Map<String, Boolean>> getView(PCollection<Event> events) {
+    return events.apply("nat view", new DetectNat()).apply(View.<String, Boolean>asMap());
```
ajvb: Is `DetectNat` now shared among each HTTPRequest monitored resource that makes use of it?

ameihm0912: no it's still one per instance, but we can do that in another PR

PR #243, opened 2019-09-10
Support fan out for iprepd output

### LINKED ISSUE #239: Support fan out push in IprepdIO
IprepdIO currently only supports submission to a single instance

This should be modified so multiple iprepd instances can be configured, and the transforms will publish to all configured instances

ajvb:
* Creates a new spec similar to kinesis for iprepd `url:key` combos
* Moves "input iprepd" to input options (used by AMO)
* Remove legacy iprepd support

Fixes #239

### DIFF
--- a/src/main/java/com/mozilla/secops/InputOptions.java
+++ b/src/main/java/com/mozilla/secops/InputOptions.java
@@ -78,4 +78,9 @@ public interface InputOptions extends PipelineOptions, PubsubOptions, GcpOptions
   Long getGenerateConfigurationTicksMaximum();
 
   void setGenerateConfigurationTicksMaximum(Long value);
+
+  @Description("Read reputation from iprepd; specify URL and API Key (supports RuntimeSecrets).")
+  String getInputIprepd();
+
+  void setInputIprepd(String value);
 }
--- a/src/main/java/com/mozilla/secops/IprepdIO.java
+++ b/src/main/java/com/mozilla/secops/IprepdIO.java
@@ -111,19 +111,29 @@ public class IprepdIO {
   /**
    * Return a new reader for reading reputation from iprepd
    *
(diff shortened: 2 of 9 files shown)


src/main/java/com/mozilla/secops/IprepdIO.java
```diff
@@ -225,6 +225,38 @@ public Reader(String url, String apiKey) {
     }
   }
 
+  /**
+   * Return {@link PTransform} to emit violations to multiple instances of iprepd
```
ameihm0912: Just a minor nit, I'm guessing docstring wise we'd want something more like `one or more instances` since it could just be one configured?

src/main/java/com/mozilla/secops/IprepdIO.java
```diff
@@ -240,7 +272,7 @@ public static Write write(String url, String apiKey, String project) {
     } catch (IOException exc) {
       throw new RuntimeException(exc.getMessage());
     }
-    return new Write(url, key, project);
+    return new Write(new String[] {String.format("{}|{}", url, key)}, project);
```
ameihm0912: Is this just to maintain compat or something? I think we should just get rid of this and only support the new specification format?

src/main/java/com/mozilla/secops/InputOptions.java
```diff
@@ -78,4 +78,14 @@
   Long getGenerateConfigurationTicksMaximum();
 
   void setGenerateConfigurationTicksMaximum(Long value);
+
+  @Description("Read reputation from iprepd; specify URL, only applicable for HTTPRequest results")
+  String getInputIprepd();
```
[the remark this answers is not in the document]
ajvb: Ah yeah, agreed. Fixed.

PR #249, opened 2019-09-18
Refit of Customs analysis pipeline

ameihm0912:
This is a follow up to removal of the generic rate limit analyzers in a2ddd19

Modify `Customs` pipeline based on existing proposal.

Adds new alert conversion transforms and publication of this new alert format. Since this is a format specific to FxA the output transform has been added to the `customs` package as we will not use the general Pubsub alert output here anymore.

Initially add simple per-source login failure analysis in conjunction with existing account creation analysis. This also adds a new analysis transform that will be enhanced to do periodic event category summaries in standard alert format.

This PR does not contain all planned analysis transforms yet, but PRing now to get this in place so we can plan to deploy it and begin enhancing it.

/cc @philbooth


ameihm0912:
See also https://github.com/mozilla/fxa/pull/2445

src/main/java/com/mozilla/secops/customs/CustomsAlert.java
```diff
@@ -0,0 +1,471 @@
...
+    DISABLE
+  }
+
+  private static HashMap<String, String> heuristicDescriptions =
+      new HashMap<String, String>() {
+        private static final long serialVersionUID = 1L;
+
+        {
+          put(
+              "account_creation_abuse",
```
ajvb: nb: Even though they are only used in two places, might be worth making these variables.

src/main/java/com/mozilla/secops/customs/CustomsAlert.java
```diff
@@ -0,0 +1,471 @@
...
+    ret.details.put("origin_alert_id", a.getAlertId());
+
+    // Default to a mid-range confidence value
+    ret.setConfidence(50);
+
+    // Set the heuristic and description fields based on the category metadata.
+    ret.setHeuristic(a.getMetadataValue("customs_category"));
+    String desc = heuristicDescriptions.get(a.getMetadataValue("customs_category"));
+    if (desc == null) {
+      ret.setHeuristicDescription("unknown");
```
ajvb: Should this be ok? Is the idea that FxA will just discard it so we don't need to worry about it if it's not found in `heuristicDescriptions`?

philbooth: Yep, the only thing FxA does with the heuristic descriptions (and most fields fwiw) is log them. Only `indicator`, `indicator_type`, `suggested_action` and `timestamp` have any impact on behaviour for now.

PR #257, opened 2019-09-30
Reusable AuthStateModel, move geo-velocity analysis into model

ameihm0912:
Move `StateModel` out of the `authprofile` package into a new `authstate` package, with the intent of making this class reusable with other pipelines.

This also moves the geo-velocity analysis that was originally part of `AuthProfile` into the state model.

### DIFF
--- a/src/main/java/com/mozilla/secops/authstate/AuthStateModel.java
+++ b/src/main/java/com/mozilla/secops/authstate/AuthStateModel.java
@@ -4,8 +4,13 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.mozilla.secops.GeoUtil;
 import com.mozilla.secops.state.StateCursor;
 import com.mozilla.secops.state.StateException;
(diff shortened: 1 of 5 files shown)


src/main/java/com/mozilla/secops/authstate/AuthStateModel.java
```diff
@@ -1,32 +1,88 @@
...
+   * After a particular source IP address has not been seen for a user in DEFAULTEXPIREAGE seconds,
+   * it will be considered "expired". However it will still be present in the stored model.
+   */
+  public static final long DEFAULTEXPIREAGE = 864000L; // 10 days
 
   private String subject;
   private Map<String, ModelEntry> entries;
 
+  /** Response to {@link AuthStateModel} GeoVelocity analysis request */
+  public static class GeoVelocityResponse {
```
ajvb: In regards to naming and function, I wonder if this should be named something more generic like `GeoChangeResponse` or similar. The change I'm working on for Authprofile will just need the `kmDistance` to consider whether the new source address is within a variable distance from the last source address.

ameihm0912: does it still require a distance/sec parameter to calculate?

ajvb: Not used at all anymore, at least for now. I still want to figure out a way to add geovelocity within the main authprofile flow.

PR #259, opened 2019-10-04
FxA velocity analysis

ameihm0912:


src/main/java/com/mozilla/secops/authstate/PruningStrategyLatest.java
```diff
@@ -0,0 +1,22 @@
+package com.mozilla.secops.authstate;
+
+import java.util.AbstractMap;
+import java.util.ArrayList;
+import java.util.HashMap;
+
+/**
+ * All entries are removed from the model with the exception of the entry with the latest timestamp.
+ */
+public class PruningStrategyLatest implements PruningStrategy {
```
ajvb: What is the need for having this different strategy for customs?

PR #270, opened 2019-10-24
improve discarded event handling

ameihm0912:
Rather than silently dropping events that do not match filters, use a
tuple tag to group these events in a separate collection. In the current
implementation, the count of events that were not OK is simply logged in
the pipeline. This could be expanded to potentially write certain types
of failures out to GCS as an example.

This also adds a new flag to drop events that have a timestamp older
than the specified period in seconds.

### DIFF
--- a/src/main/java/com/mozilla/secops/input/InputDiscardSummary.java
+++ b/src/main/java/com/mozilla/secops/input/InputDiscardSummary.java
@@ -1,0 +1,93 @@
+package com.mozilla.secops.input;
+
+import com.mozilla.secops.parser.Event;
+import org.apache.beam.sdk.transforms.Count;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.transforms.PTransform;
+import org.apache.beam.sdk.transforms.ParDo;
+import org.apache.beam.sdk.transforms.windowing.FixedWindows;
+import org.apache.beam.sdk.transforms.windowing.Window;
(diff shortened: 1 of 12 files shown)


src/main/java/com/mozilla/secops/input/InputDiscardSummary.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.beam.sdk.transforms.windowing.Window;
+import org.apache.beam.sdk.values.KV;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PDone;
+import org.joda.time.Duration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Summarize input discard events */
+public class InputDiscardSummary extends PTransform<PCollection<Event>, PDone> {
```
ajvb: In particular for `TIMESTAMP_TOO_OLD`, it might be worth considering sending metrics via `org.apache.beam.sdk.metrics.Metrics`. As well, either way, it would be nice to know within the summary what event types are hitting the `TIMESTAMP_TOO_OLD`.

ameihm0912: yeah good points; I am going to hold off on merging this to address what you mentioned and I also think there might be some improvements we can still make to this PR

ameihm0912:
Superseded by #274

PR #271, opened 2019-10-24
Integrate Minfraud into authprofile

ajvb:
* Implements new flow for authprofile using minfraud insights data
* Removes use of slack feature flag within identity manager
* Adds minfraud data into alert metadata


src/main/java/com/mozilla/secops/authprofile/AuthProfile.java
```diff
@@ -630,46 +597,78 @@ public void processElement(ProcessContext c) throws StateException {
...
+                  log.info(
+                      "{}: escalating alert criteria for new source outside of allowed distance from last login: {} {}",
+                      userIdentity,
+                      e.getNormalized().getSubjectUser(),
+                      e.getNormalized().getSourceAddress());
+                  a.setSeverity(Alert.AlertSeverity.WARNING);
+                  buildAlertPayload(
+                      e,
+                      a,
+                      "This occurred from a source address unknown to the system and outside of the allowed range from your last auth event.");
```
ameihm0912: nit: `auth` -> `authentication` ?

src/main/java/com/mozilla/secops/authprofile/AuthProfile.java
```diff
@@ -630,46 +597,78 @@ public void processElement(ProcessContext c) throws StateException {
...
+                      "{}: escalating alert criteria for new source inside of allowed distance from last login: {} {}",
+                      userIdentity,
+                      e.getNormalized().getSubjectUser(),
+                      e.getNormalized().getSourceAddress());
+                  a.setSeverity(Alert.AlertSeverity.WARNING);
+                  onlyNotify = true;
+                  buildAlertPayload(
+                      e,
+                      a,
+                      "This occurred from a source address unknown to the system, but near your last auth event.");
```
ameihm0912: Same thing here.

src/main/java/com/mozilla/secops/parser/SourcePayloadBase.java
```diff
@@ -60,6 +68,19 @@ public void setSourceAddress(String sourceAddress, ParserState state, Normalized
       n.setSourceAddressLatitude(sourceAddressLatitude);
       n.setSourceAddressLongitude(sourceAddressLongitude);
     }
+
+    if (n != null && (n.isOfType(Normalized.Type.AUTH) || n.isOfType(Normalized.Type.AUTH))) {
+      InsightsResponse ir = state.getParser().insights(sourceAddress);
```
ameihm0912: I think we need to change this part here, I definitely agree this is the best way to do this architecturally but I think this will burn through our quota in no time (e.g., `Cloudtrail`, `Taskcluster`, and `GcpAudit` are `SourcePayloadBase`). This will also make the parser IO bound on the API calls and could slow things down too much.

We keep a seen address list in the state analyze `ParDo`, partly so we are not doing state checks for duplicates that all fall into the same pane. I'm wondering if it would be more appropriate to move the insights enrichment out of the parser; you could perhaps just add a function to `SourcePayloadBase` along the lines of `insightsEnrichment` that will add the required information to the event on demand, and we can just call it when necessary in the `DoFn`?

src/main/java/com/mozilla/secops/Minfraud.java
```diff
@@ -0,0 +1,109 @@
...
+  private final Logger log;
+  private WebServiceClient mfClient;
+
+  /**
+   * Get Insights response from Minfraud using an IP address and an optional email address
+   *
+   * <p>See Maxmind's minfraud documentation for more details.
+   *
+   * @param ip IP address string (required)
+   * @param email email address string (optional)
```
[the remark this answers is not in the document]
ajvb: Discussed on slack, but yeah, IP is required. Can send a bogus IP though.

src/main/java/com/mozilla/secops/parser/SourcePayloadBase.java
ajvb: Totally agree. Moved into `Normalized.insightsEnrichment()`, called within AuthProfile.

src/main/java/com/mozilla/secops/authprofile/AuthProfile.java
```diff
@@ -452,67 +454,35 @@ private String getEntryKey(String ipAddr) {
...
 
       String dnote = identity.getEmailNotifyDirect(idmanager.getDefaultNotification());
       if (dnote != null) {
-        log.info(
-            "{}: adding direct email notification metadata route to {}",
-            a.getMetadataValue("identity_key"),
-            dnote);
-        a.addMetadata("notify_email_direct", dnote);
+        if (onlyNotify) {
+          a.addMetadata("notify_email_direct", identityKey);
```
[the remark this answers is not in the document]
ajvb: Simplified `Identity`. There now exists a `notify` and an `alert` section for each identity. Includes an optional `email` field and a required `method` field. Removed the defaults, feature flags, and fragments. Also makes the API for `Identity` much simpler.

src/main/java/com/mozilla/secops/authprofile/AuthProfile.java
```diff
@@ -597,11 +566,12 @@ public void processElement(ProcessContext c) throws StateException {
           // logic here
           a.addMetadata(AlertIO.ALERTIO_IGNORE_EVENT, "true");
           buildAlertSummary(e, a);
-          buildAlertPayload(e, a);
+          buildAlertPayload(e, a, null);
```
ameihm0912: By passing `null` here does this mean the alert message will indicate an untracked identity? I don't think that will always be the case in this code block?

ajvb: Yes it does. Fixing.

src/main/java/com/mozilla/secops/authprofile/AuthProfile.java
```diff
@@ -631,46 +610,78 @@ public void processElement(ProcessContext c) throws StateException {
...
+                  "{}: escalating alert criteria for new source from hosting provider: {} {}",
+                  userIdentity,
+                  e.getNormalized().getSubjectUser(),
+                  e.getNormalized().getSourceAddress());
+              a.setSeverity(Alert.AlertSeverity.WARNING);
+              buildAlertPayload(
+                  e,
+                  a,
+                  "This occurred from a source address unknown to the system and marked as from a hosting provider.");
+            } else {
```
ameihm0912: What would happen here if the minFraud lookup failed, but GeoIP was successful? It seems like in this case, we'd want to basically just ignore GeoIP and handle it as a new address since we do not know the status of the address (hosting, etc)?

ajvb: Totally, fixing.

ajvb:
@ameihm0912 Spent awhile trying to mock `Minfraud` to no avail. Only way I could see to do so successfully was to pull out the core of `Authprofile.StateAnalyze.processInput()` into it's own function (similar to `createBaseAlert`), but that seemed like a gnarly change for one test. So, for now, added an option to ignore Minfraud to allow unit testing of geo distance.

PR #274, opened 2019-11-04
improve old event handling

ameihm0912:
Optionally drop events with a timestamp a certain number of seconds in
the past; if events are dropped, a custom metric to indicate the input
transform path is incremented.


ameihm0912:
@ajvb this is intended to replace #270 with a less impactful change.

PR #289, opened 2019-12-05
Improvements to Customs analysis

### LINKED ISSUE #286: Refine single selector customs transforms with variance component
Several customs transforms should be refined with a variance component including:

* `CustomsPasswordResetAbuse`
* `CustomsAccountCreation`
* `SourceLoginFailure`

to work around issues associated with large residential/mobile NAT pools.

ameihm0912:
This is a fairly substantial rework of how Customs analysis occurs.

Previously, most analysis transforms were responsible for all aspects of consuming and processing the parsed event stream. In a lot of cases these transforms were duplicating effort and this also made it a bit harder to add shared functionality to the analysis as required in #286.

This modifies customs so all transforms expect a new `CustomsFeatures` element which contains  a similar raw event set keyed as required as they would previously expect, in addition to various other properties of the event set applied as statistics generated by a `CombineFn` on the consumed pane.

In this model, `CustomsFeatures` can be enhanced as needed and these new statistics/properties become available to the analysis transforms with minimal refactoring of the pipeline graph.

In addition to this primary change, this PR also:

* Makes the pipeline option names more consistent
* Removes the `PTransform` that previously encapsulated the account creation heuristics
* Introduces the variance index as a property, described in #286 

This also unifies the windowing strategy for efficiency. This is expected to change however, and future work here will involve replacing the default trigger which is currently being used.

Closes #286


ameihm0912:
Issue #290 created as a follow up to use of the default trigger here.

PR #291, opened 2019-12-06
Customs enhancements

### LINKED ISSUE #290: Replace default trigger in Customs
#289 modifies the windowing strategy used in Customs, but makes use of a default trigger.

This should be further enhanced with early pane firings, which could include potentially dealing with suppressing the on-time pane if needed to prevent duplicates.

ameihm0912:
A few changes as a follow up to #289.

* When storing a raw event list during feature extraction, only store events we need to look at, but keep counters for everything. This reduces collection size and the need to pass around data we will never look at.
* Add early pane firings for windowed collections so we get early results.
* Don't force recalculation during feature merge, and only calculate once results are extracted from the step
* Filter high volume "ping" events that greatly increase the cost of downstream shuffle steps.

Closes #290


ameihm0912:
> This means only call recalculate in `CustomsFeaturesCombiner`, right?

Yeah, basically just on output extraction and not as part of accumulator merge.

PR #302, opened 2020-01-15
PostProcessing pipeline

### LINKED ISSUE #301: Alert post-processing pipeline
New pipeline, intended to:

* Use standard composite input (read input topic containing `Alert` JSON strings)
* Use standard composite output (e.g., BQ)

Initially it would be ideal for this to consume a topic containing `Alert` JSON strings, parse them into `Alert` objects, and push these objects into an analysis step.

The analysis step will look for metadata fields that for example match a given regex, and where a match occurs a new alert will be generated. This alert will contain a metadata field containing the ID of the source alert which it matched on.

This new pipeline will also be useful for future correlation between pipeline alerting output.

ajvb:
Fixes #301

One way in which this currently differs from #301, is that instead of using or support regex, it's exact match based off of metadata key names for IP and email addresses.

The `WatchlistAnalyze` step within `PostProcessing` fetches email and IP address from the datastore namespace `watchlist`. It then looks for matches in incoming Alerts using specified metadata keys, and if it finds a match it will output an alert. Multiple alerts can be created off of a single incoming alert if there are multiple matches (likely IP and Email matching).

Missing pieces:
* Currently no interface for adding watchlist entries to Datastore. This should probably be done via the slack bot.
* Currently no process for escalating alerts. We should discuss this within our team meeting (or within the PR if desired)


src/main/java/com/mozilla/secops/Watchlist.java
```diff
@@ -0,0 +1,213 @@
...
+      s = emailState;
+    } else if (type.equals(watchlistIpKind)) {
+      s = ipState;
+    } else {
+      return null;
+    }
+
+    WatchlistEntry[] entries;
+    try {
+      s.initialize();
```
ameihm0912: Should this be in the constructor in case `getWatchedObjects` is called more than once using the same object?

src/main/java/com/mozilla/secops/Watchlist.java
```diff
@@ -0,0 +1,213 @@
...
+      entries = s.getAll(WatchlistEntry.class);
```
ameihm0912: Might be lower priority right now but I am wondering if it's make sense to use query filters here instead of fetching everything.

src/main/java/com/mozilla/secops/Watchlist.java
```diff
@@ -0,0 +1,213 @@
...
+    } catch (StateException exc) {
+      log.error("Error getting all watched emails: {}", exc.getMessage());
```
ameihm0912: Hardcoded reference to `emails` here when it could be `ip` type

src/main/java/com/mozilla/secops/Watchlist.java
```diff
@@ -0,0 +1,213 @@
...
+  /**
+   * Returns watched ip addresses
+   *
+   * @return Array of WatchlistEntry of type "ip"
+   */
+  public WatchlistEntry[] getWatchedIPs() {
+    return getWatchedObjects(watchlistIpKind);
+  }
+
+  /** Closes state interfaces to datastore */
```
ameihm0912: It might be worth adding a note in the doc here that this must be called when we are done using the object, just for consistency.

src/main/java/com/mozilla/secops/postprocessing/PostProcessing.java
```diff
@@ -0,0 +1,272 @@
...
+        }
+        for (Watchlist.WatchlistEntry entry : entries) {
+          if (v.equals(entry.getObject())) {
+            Alert a = new Alert();
+            a.setCategory("postprocessing");
+            a.addMetadata("category", "watchlist");
+            a.setSummary(
+                String.format(
+                    "matched watchlist object found in alert %s", inputAlert.getAlertId()));
+            a.setSeverity(entry.getSeverity());
```
ameihm0912: Should this be based on the incoming alert or always for example be critical/etc?

src/main/java/com/mozilla/secops/postprocessing/PostProcessing.java
```diff
@@ -0,0 +1,272 @@
...
+
+            a.addMetadata("input_alert", inputAlert.getAlertId().toString());
```
ameihm0912: It might be good to store a metadata value containing just the source alert ID for metrics/SQL join purposes.

Also need to be careful about PII ending up in here, and ignore this field all the time in the required view.

src/main/java/com/mozilla/secops/Watchlist.java
ajvb: I'm totally open to it, I'm just not sure what the filter would exclude at the moment?

src/main/java/com/mozilla/secops/postprocessing/PostProcessing.java
ajvb: Well it's based off of the watchlist entry. My thought was that severity would be used to determine how the alert is routed to us (i.e. pagerduty vs. slack) and that we would usually want to determine that while creating the watchlist entry.

src/main/java/com/mozilla/secops/postprocessing/PostProcessing.java
ajvb: changing the name from `input_alert` to `source_alert` for consistency, but yeah that is exactly what this line is doing (unless I'm misunderstanding your request?).

src/main/java/com/mozilla/secops/Watchlist.java
ameihm0912: I was thinking along the lines of a "does this particular value exist in the watchlist" query vs. grabbing them all and then iterating? Right now with this approach, since we fetch everything for each element in `processElement` we are downloading the entire watchlist contents for each alert.

An alternative would be to just fetch the watchlist contents once in `@setup` for the `DoFn`, but this has the limitation it wouldn't allow for updates once running.

src/main/java/com/mozilla/secops/postprocessing/PostProcessing.java
ameihm0912: Ah OK I see here now, so the severity value is actually stored with the watchlist entry. Sounds good 👍

src/main/java/com/mozilla/secops/Watchlist.java
ajvb: Oh totally. Yeah that makes way more sense. Switching to that.

src/main/java/com/mozilla/secops/Watchlist.java
```diff
@@ -0,0 +1,249 @@
...
+      entry = s.get(obj, WatchlistEntry.class);
+    } catch (StateException exc) {
+      log.error("Error watchlist entry of type {}: {}", type, exc.getMessage());
+      return null;
+    }
+
+    return entry;
+  }
+
+  private WatchlistEntry[] getWatchedObjects(String type) {
```
ameihm0912: Did we leave this in just for use in tests? (OK either way just want to make sure it wasn't left in accidentally)

src/main/java/com/mozilla/secops/state/MemcachedStateCursor.java
```diff
@@ -21,6 +21,10 @@ public String getObject(String s) throws StateException {
     }
   }
 
+  public String[] getAllObjects() throws StateException {
+    throw new StateException("getAll not implemented for memcached");
```
ameihm0912: Should this be `getAllObjects`?

src/main/java/com/mozilla/secops/Watchlist.java
ajvb: No, this is what `getWatchedEmails` and `getWatchedIPs` use

ajvb: Oh, my bad, yes this is just used by tests as of now.

PR #305, opened 2020-01-21
Update httpclient, add iprepd write timeouts

ameihm0912:
Updates httpclient to 4.5.11.

This also adds connection, socket, and connection manager timeout values to `IprepdIO.WriteFn`. This should help prevent cases where the output step hangs if iprepd does not respond to a particular `PUT` request correctly or in a timely manner.

### DIFF
--- a/src/main/java/com/mozilla/secops/IprepdIO.java
+++ b/src/main/java/com/mozilla/secops/IprepdIO.java
@@ -23,6 +23,7 @@ import org.apache.beam.sdk.values.PDone;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpResponse;
 import org.apache.http.client.HttpClient;
+import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpPut;
 import org.apache.http.entity.StringEntity;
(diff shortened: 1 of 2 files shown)


src/main/java/com/mozilla/secops/IprepdIO.java
```diff
@@ -330,6 +331,15 @@ public PDone expand(PCollection<String> input) {
     private String[] iprepdSpecs;
     private String project;
 
+    /** Writer initial connection timeout */
+    public final int WRITER_TIMEOUT_CONNECTION = 5000;
```
kkleemola: This seems sane but I wonder if this should be configurable with 5000 as a default.

src/main/java/com/mozilla/secops/IprepdIO.java
```diff
@@ -330,6 +331,15 @@ public PDone expand(PCollection<String> input) {
...
+
+    /** Writer connection manager connection request timeout */
+    public final int WRITER_TIMEOUT_CONNECTION_REQUEST = 5000;
```
kkleemola: Same as above.

src/main/java/com/mozilla/secops/IprepdIO.java
```diff
@@ -330,6 +331,15 @@ public PDone expand(PCollection<String> input) {
...
+
+    /** Writer socket timeout */
+    public final int WRITER_TIMEOUT_SOCKET = 5000;
```
kkleemola: Same as above.

src/main/java/com/mozilla/secops/IprepdIO.java
ameihm0912: Created #306 for this

PR #307, opened 2020-01-24
Use CircleCI

ameihm0912:
Migrates CI to Circle and adds image push to Dockerhub.

The docker image here was previously used only for development purposes.

With these changes, two images can now be built.

`foxsec-pipeline-base` (`Dockerfile-base`) is now the development image, and contains the required toolchain and emulators for development and running tests.

`foxsec-pipeline` (`Dockerfile-complete`) is a complete image that also contains the compiled code, and is pushed to DH.


docker-entrypoint.sh
```diff
@@ -2,18 +2,23 @@
...
+	setsid nohup gcloud beta emulators datastore start >/var/log/datastore.out 2>&1 &
+	setsid nohup gcloud beta emulators pubsub start >/var/log/pubsub.out 2>&1 &
+	while ! nc -z localhost 8081; do sleep 0.1; done
+	while ! nc -z localhost 8085; do sleep 0.1; done
 
-/usr/bin/memcached -u root -d
-/usr/bin/redis-server --daemonize yes
-while ! nc -z localhost 11211; do sleep 0.1; done
-while ! nc -z localhost 6379; do sleep 0.1; done
+	/usr/bin/memcached -u root -d
```
jasonthomas: I am not sure why this is a dark green. Is it a tab?

.circleci/config.yml
```diff
@@ -0,0 +1,27 @@
...
+          name: Build complete pipeline image
+          command: docker build -f Dockerfile-complete -t foxsec-pipeline:latest .
+      - run:
+          name: Test complete pipeline image
+          command: docker run -ti --rm foxsec-pipeline:latest mvn test
+      - run:
+          name: Push to Dockerhub
+          command: |
+            if [[ "${CIRCLE_BRANCH}" == "master" || -n "${CIRCLE_TAG}" ]]; then bin/docker_push.sh; fi
+workflows:
```
jasonthomas: I think we should split the workflow here so that we only push to dockerhub when master and tags are updated. https://github.com/mozilla/messaging-system-personalization-experiment-1-numbermuncher/blob/dev/.circleci/config.yml has an example.

docker-entrypoint.sh
ameihm0912: It is yeah, the rest of the shell scripts use tab indentation so just left it as is.

.circleci/config.yml
ameihm0912: Discussed with @jasonthomas we are just going to leave this as is for now since it's conditional on the branch/tag in the shell command.

PR #315, opened 2020-02-11
Unit test for slackbot http handler

ajvb:
Adds a unit test (that requires the pubsub emulator) for the SlackbotHTTP handler. It tests forwarding along a slash command and interaction callback as well as doing nothing when lacking a proper slack signature.


kkleemola:
:100:

PR #320, opened 2020-02-13
ingestion and alert source correlation

ameihm0912:


src/main/java/com/mozilla/secops/alert/AlertFormatter.java
```diff
@@ -21,13 +17,51 @@
   private String maxmindIspDbPath;
   private GeoIP geoip;
 
+  /**
+   * SimpleFunction for conversion of {@link Alert} objects to String
+   *
+   * <p>Intended for use with MapElements.
+   */
+  public static class AlertConversion extends SimpleFunction<Alert, String> {
```
ajvb: Could this be called `ConvertToString` or similar?

When I read `.apply("alert conversion", MapElements.via(new AlertFormatter.AlertConversion()))` below, my first thought is "covert to what?".

src/main/java/com/mozilla/secops/SourceCorrelation.java
```diff
@@ -0,0 +1,338 @@
...
+ *
+ * <p>Correlates observed source addresses extracted from normalized event field with source
+ * addresses that are associated with alerts.
+ */
+public class SourceCorrelation {
+  /**
+   * SourceData is an intermediate format used to store information about a given source address
+   * observed in the ingestion or alert stream.
+   *
+   * <p>{@link AvroCoder} is used to guarantee certain uniqueness properties that are needed during
```
ajvb: Could this be expanded to further explain what the uniqueness properties are and why AvroCoder? Or even link to some documentation that would explain it?

src/main/java/com/mozilla/secops/SourceCorrelation.java
```diff
@@ -0,0 +1,338 @@
...
+   * based on distinct source address, and are compared with source address and ISP information
+   * grouped in a similar way from the ingestion event stream.
+   *
+   * <p>If at least a certain number of addresses are seen associated with a given ISP, and the
+   * percentage of these addresses that generated an alert in a fixed window exceeds the configured
+   * value, an alert is generated.
+   *
+   * <p>Currently uses a hardcoded fixed window value of 6 hours.
+   */
+  public static class SourceCorrelator
```
ajvb: Should `SourceCorrelator` implement `DocumentingTransform` and have a transform doc?

ameihm0912: Yes I think it should, missed that -- I will add that

src/main/java/com/mozilla/secops/SourceCorrelation.java
ameihm0912: Sounds good yeah, will add that

src/main/java/com/mozilla/secops/alert/AlertFormatter.java
ameihm0912: I think that makes sense too, will modify so it's clearer

ameihm0912:
> Looks great, just left some documentation/naming comments.
> 
> One thing I was curious about while reading through this was whether we have data showing that a streaming method of implementing this heuristic is preferred over using a batch job? Seeing the `Window.<SourceData>into(FixedWindows.of(Duration.standardHours(6))) ` and considering how easy it could be for an ISP to get unalerted upon if the requests come in just a bit too slow to get detected, could a batch job make more sense?

What sort of batch job are you thinking here? This depends on enriched `Event` collections that we don't store, and depends on metrics pulled from good clients we'd never alert on. To implement as a batch job we would have to pull in historical event data from potentially different data sources, it's doable and more resistant to things like a pipeline restart but would likely be quite a bit more complicated, unless we want to get into summarizing and persisting known-good data?

ajvb:
I think I got ahead of myself with talking about a batch job. The question I'm wondering about is whether 6 (or similar) hours is usually long enough to get the 90% of 250 (or similar) unique IP's required to alert? Or, are we usually getting to "90% of a large enough number of unique IP's from a single ISP" over the course of a longer period of time?


> What sort of batch job are you thinking here? This depends on enriched Event collections that we don't store, and depends on metrics pulled from good clients we'd never alert on. To implement as a batch job we would have to pull in historical event data from potentially different data sources, it's doable and more resistant to things like a pipeline restart but would likely be quite a bit more complicated, unless we want to get into summarizing and persisting known-good data?

I do agree that it would be quite a bit more complicated, but would we have to persist good clients in a way much different from how this pipeline is holding all of the `SourceData` events? My thought with a batch job is that it could perform the same analysis in the same way, just using a longer time period of data. It would require persisting all of the `SourceData` events in a way we haven't done yet in the pipeline, but I would think the analysis implementation would be pretty much the same as this pipeline, just that it could look at the last 14 days of `SourceData` events or something.

ameihm0912:
> It would require persisting all of the `SourceData` events in a way we haven't done yet in the pipeline

Discussed this offline and we will continue with this as is, but also start laying some groundwork so we can take this approach (issue #321).

PR #322, opened 2020-02-19
Upgrade jslack to get new version of okhttp

ajvb:
Need okhttp version 3.13 or higher so that it won't try and use TLS 1.0
or 1.1 with slack:
https://api.slack.com/changelog/2019-07-deprecate-early-tls-versions


ajvb:
jslack version 1.8.1 uses version 3.14.1 of okhttp3 - https://github.com/slackapi/java-slack-sdk/blob/jslack-1.8.1/pom.xml#L30

PR #326, opened 2020-02-24
Cache download go/contrib deps

ajvb:
Follows advice in https://medium.com/@petomalina/using-go-mod-download-to-speed-up-golang-docker-builds-707591336888 to cache go modules. Unless a `go.sum` or `go.mod` file changes, everything before `COPY . /root/project` will stay cached.


Dockerfile-complete
```diff
@@ -5,17 +5,39 @@
...
+COPY contrib/auth0pull/go.sum /root/project/contrib/auth0pull/
+COPY contrib/auth0pull/go.mod /root/project/contrib/auth0pull/
+COPY contrib/cloudtrail-streamer/go.sum /root/project/contrib/cloudtrail-streamer/
+COPY contrib/cloudtrail-streamer/go.mod /root/project/contrib/cloudtrail-streamer/
+COPY contrib/duopull/go.sum /root/project/contrib/duopull/
+COPY contrib/duopull/go.mod /root/project/contrib/duopull/
+COPY contrib/slackbot-background/go.sum /root/project/contrib/slackbot-background/
+COPY contrib/slackbot-background/go.mod /root/project/contrib/slackbot-background/
+COPY contrib/slackbot-http/go.sum /root/project/contrib/slackbot-http/
+COPY contrib/slackbot-http/go.mod /root/project/contrib/slackbot-http/
```
ameihm0912: I guess there isn't much we can do here to avoid creation of all the extra layers?

ajvb: @ameihm0912 Was able to cut it in half by using `COPY contrib/go.* /root/project/contrib/`, but yeah from my understanding of `COPY` that's about the best that can be done.

PR #327, opened 2020-02-25
alert suppression for critical object analyze

ameihm0912:


ameihm0912:
Merged but wasn't closed; closing manually

PR #332, opened 2020-03-02
Endpoint Sequence timing heuristic

kkleemola:
Adds support for request pattern timing heuristics
i.e. doing x within delta of y will create an alert.

Note: this operates on fixed windows for simplicity (we need to sort events by event timestamp to do the analysis so having smaller windows will be beneficial).


src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
```diff
@@ -972,6 +975,328 @@ public void processElement(ProcessContext c, BoundedWindow w) {
...
+                          String method = n.getRequestMethod();
+                          String path = n.getUrlRequestPath();
+                          if (sourceAddress == null || method == null || path == null) {
+                            return;
+                          }
+                          Map<String, Boolean> nv = c.sideInput(natView);
+                          Boolean isNat = nv.get(sourceAddress);
+                          if (isNat != null && isNat) {
+                            log.info(
+                                "{}: detectnat: skipping result emission for {}",
```
ameihm0912: I'm wondering if it would be more appropriate to move this check to later on, like right before we emit an alert. The location of this check now means we will emit this log line for every single event associated with the source.

Also the side input lookup is very expensive, so it's better to try to apply this to the smallest collection size possible (e.g., after we have done all the grouping and are about to send the alert, do the check vs. a side input lookup for every event).

src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
```diff
@@ -972,6 +975,328 @@ public void processElement(ProcessContext c, BoundedWindow w) {
...
+                                w.toString(),
+                                sourceAddress);
+                            return;
+                          }
+                          // only output events if they belong to one of our sequences
+                          if (belongsToSequence(method, path)) {
```
ameihm0912: Do we care if there are events in between? For example, if we are looking at request type A and then D, do we care if we have B and C between them? Maybe not for the purposes of this but I was wondering if it'd make sense to somehow reflect the number/types of requests in between or not.

src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
kkleemola: Done.

src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
kkleemola: I'm not sure it should change whether we alert or not. It might be interesting to know if there's other events in between (maybe to help detect nats?)
The trade off is then we can't discard all those events before sorting the events by timestamp.

I suppose if we are sorting all events by ip in a window there may also be additional heuristics we could write based on it, like do we typically see less deviation in time between requests for bots (i.e. not just looking at the request patterns we've decided we care about?)

PR #338, opened 2020-03-16
Guardduty Alert Handling

ajvb:
* Changes gatekeeper to have high and low severity alerts via their metadata
* Adds `bugzilla-alert-manager` - a cloud function that reads alerts from pubsub and manages daily bugzilla bugs with the alerts.

In `bugzilla-alert-manager`, a Pagerduty Schedule is used as the mechanism to hold the rotation schedule. This allows us to use all of the nice scheduling features in Pagerduty while still doing the actual alert handling ourselves.


contrib/bugzilla-alert-manager/bugzilla.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func CreateBugJSONFromAlerts(alerts []*common.Alert) ([]byte, error) {
+	assignedTo, err := userOnTicketDuty()
+	if err != nil {
+		log.Errorf("Couldn't get user on ticket duty: %s", err)
+		return nil, err
+	}
+
+	summary := fmt.Sprintf("Guardduty Alerts for %s", time.Now().Format("2012-11-01"))
```
ameihm0912: I'm wondering about hard coding alert types in here. Should we derive this information from the alert itself and not assume it is a GD alert? It would be good to use this code for other things as well.

contrib/bugzilla-alert-manager/bugzilla.go
```diff
@@ -0,0 +1,78 @@
...
+		"task",
+		[]string{"mozilla-employee-confidential", "cloud-services-security"},
+		true,
+		globals.bugzillaApiKey,
+	}
+}
+
+func CreateBugJSONFromAlerts(alerts []*common.Alert) ([]byte, error) {
+	assignedTo, err := userOnTicketDuty()
+	if err != nil {
```
ameihm0912: If `err` is non-nil here it looks like we exit. Could we change this so instead we still create the bug, but maybe it's just unassigned?

contrib/bugzilla-alert-manager/bugzilla.go
```diff
@@ -0,0 +1,78 @@
...
+		log.Errorf("Couldn't get user on ticket duty: %s", err)
+		return nil, err
+	}
+
+	summary := fmt.Sprintf("Guardduty Alerts for %s", time.Now().Format("2012-11-01"))
+
+	// TODO - improve
+	bugText := "## Guardduty Alerts\n---\n"
```
ameihm0912: Similar to previous comment here; I think this function should be generic and not make assumptions about alert type, instead we should derive all of that from the alert payload and metadata.

contrib/bugzilla-alert-manager/bugzilla.go
```diff
@@ -0,0 +1,78 @@
...
+	for _, alert := range alerts {
+		bugText = bugText + fmt.Sprintf("%s\n---\n", alert.PrettyPrint())
+	}
+	b := NewCreateBug(summary, bugText, assignedTo)
+	return json.Marshal(b)
+}
+
+func bugCreatedToday(bug *common.BugzillaBug) bool {
+	ny, nm, nd := time.Now().Date()
+	y, m, d := bug.CreatedAt.Date()
```
ameihm0912: Is this dependent at all on the timezone being used on the account used for BMO integration or will `bug.CreatedAt.Date()` always return UTC?

contrib/bugzilla-alert-manager/manager.go
```diff
@@ -0,0 +1,199 @@
...
+		return
+	}
+	mozlogrus.Enable("bugzilla-alert-manager")
+	InitConfig()
+
+	PROJECT_ID := os.Getenv("GCP_PROJECT")
+	var err error
+	DB, err = common.NewDBClient(context.Background(), PROJECT_ID)
+	if err != nil {
+		log.Errorf("Error creating db client: %s", err)
```
ameihm0912: Should this error be fatal?

contrib/bugzilla-alert-manager/manager.go
```diff
@@ -0,0 +1,199 @@
...
+	users, err := globals.pagerdutyClient.ListOnCallUsers(config.PagerdutyTicketDutyScheduleId, pagerduty.ListOnCallUsersOptions{})
+	if err != nil {
+		log.Errorf("Error getting pagerduty schedule: %s", err)
+		return "", err
+	}
+	if len(users) == 0 {
+		log.Errorf("No oncall user found for schedule %s!", config.PagerdutyTicketDutyScheduleId)
+		return "", err
+	}
+	return users[0].Email, nil
```
ameihm0912: It's probably fine but just wondering if we can always assume the PD email will match the BMO account email? If that is not the case we may be able to leverage the identity manager configuration to resolve the correct address (with some added fields there).

contrib/bugzilla-alert-manager/manager.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+func addAlertsToBug(bug *common.BugzillaBug, alerts []*common.Alert) error {
+	comment := createCommentFromAlerts(alerts)
+	commentJson, err := json.Marshal(comment)
+	if err != nil {
+		return err
+	}
+	resp, err := http.Post(
```
ameihm0912: Not directly applicable to this but more generally I am wondering what happens if `err` is non-nil throughout this flow. Would there be a way to tell which alerts should have been part of the bug if it fails. Would we only know things have failed by looking at the error log? Does this need to be monitored?

contrib/bugzilla-alert-manager/manager.go
```diff
@@ -0,0 +1,199 @@
...
+		}
+	}
+
+	if len(lowSeverityAlerts) == 0 {
+		return nil
+	}
+
+	// If more alert types are processed by this cf, then we will want to change this
+	// part around to handle which kind of bugzilla bug it should look for.
+	bugs, err := DB.GetBugzillaBugsByKind(ctx, GUARDDUTY_KIND)
```
ameihm0912: I'm just wondering why Datastore is needed here? Could we just search for the bug using a whiteboard identifier or some other search parameters and determine this without needing to keep state on the pipeline side?

ajvb:
@ameihm0912 This all sounds good to me and will add fixes.

The part I'm worried about is using Bugzilla search over just storing the ID in Datastore. Maybe it is just a biased preference over something I'm more comfortable, but I see the Bugzilla search API and making sure the Bugzilla permissions work as much more finicky than just storing the ID in Datastore. As well, we could make any modifications we want to the bug without worrying about if that will mess up our ability to fetch it. I wouldn't want to introduce a new dependency to solve this, but since the use of Datastore has been well tested in CF's and "just works" it feels like the better move at this point.

Are there further reasons to use Bugzilla's search API that I'm missing?

ameihm0912:
> Are there further reasons to use Bugzilla's search API that I'm missing?

I don't think so, basically what I am trying to avoid here is introducing a dependency if we don't need to and there are some cache invalidation concerns here. From what I gather what we want to do is see if a bug already exists, if not create it and if so update it. It seems as though just asking BMO that might be a better route since it's really the authoritative source on that. The Datastore entry is more secondary in the sense if the bug is resolved or otherwise modified in some way the Datastore entry is now stale and may not be usable depending on what happened to the bug.

I read the Datastore component here as being almost like a cache/hint but missing invalidation, which would require a query anyway to confirm it's still accurate so at that point why not just ask BMO directly?

ajvb:
@ameihm0912 Ok, that makes sense and I agree. I was looking at in that I was more concerned about it accidentally making more than 1 bug per day, but as you said it could also easily get into a broken state if the bug was resolved or modified in a way in which the alert manager could no longer post comments but also wouldn't make a new bug.

contrib/bugzilla-alert-manager/manager.go
ajvb: Good catch! I checked for our team and they all match up as of now.

ajvb: If this turns into a problem at all, we can use the People API to get folks "bugzilla email"

contrib/bugzilla-alert-manager/bugzilla.go
ajvb: Nope! It will always use UTC. As well, BMO's "creation_time" returns as a UTC datetime

PR #339, opened 2020-03-18
alert summaries framework in PostProcessing

ameihm0912:
Update `PostProcessing` with new transform that does alert summaries across sliding windows, generating notifications if certain thresholds (either increase or decrease) are observed across the window.

Thresholds can be specified as either global (all alerts), for a specific monitored resource, or further for a specific monitored resource on a certain alert category and subcategory. See the documentation for `AlertSummary` for details on the configuration format.

Windows are currently:
* Compare thresholds across adjacent 15 minute windows
* Compare thresholds across adjacent 1 hour windows

When a threshold is violated an `Alert` is generated that includes a notification entry for the warning address.

In addition to this change, this PR also adds a new `setSubcategory` method to the `Alert` class so we can standardize on one subcategory field. Previously some pipelines were using different metadata keys for subcategory. This required having logic in this pipeline to identify which type of alert it was so we could select the right metadata field. This has been changed so we just have one, and a compatibility layer has been added to maintain processing of the old event format.


ameihm0912:
> lgtm.
> 
> What do the `s` and `l` and `sblock` and `lblock` mean? small and large?

Yeah not the best choice or most obvious, I will fix some of that up before it's merged.

PR #343, opened 2020-03-23
Gatekeeper / BMO Alert Manager fixes

ajvb:
Collection of fixes:

* Fix "url_to_finding" creation in gatekeeper
* Add `.gcloudignore` for deploying the alert manager
* Refactor to support 1 alert at a time from pubsub
* Add additional logging
* Fixes for bugzilla requests: custom UA, set all of the correct headers, move api key to header auth
* Add and use testing config file to make sure the config parsing is included in unit tests.


contrib/bugzilla-alert-manager/manager.go
```diff
@@ -102,61 +103,54 @@ func BugzillaAlertManager(ctx context.Context, psmsg pubsub.Message) error {
...
-		if err != nil {
-			contextLogger.Errorf("Error from bug searching: %s", err)
-			continue
-		}
-		if len(searchResp.Bugs) != 0 {
+	searchValues := url.Values{}
+	searchValues.Add("whiteboard", alert.Category)
+	searchResp, err := globals.bugzillaClient.SearchBugs(searchValues)
+	if err != nil {
+		contextLogger.Errorf("Error from bug searching: %s", err)
```
ameihm0912: Is the lack of return here intentional so we will always create a new one if the search fails type thing?

contrib/common/bugzilla.go
```diff
@@ -63,22 +75,38 @@ func (bc *BugzillaClient) CreateBugFromAlerts(assignedTo, category string, alert
...
+		return 0, err
+	}
+	req.Header.Add("Content-Type", "application/json")
+
+	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
 		log.Errorf("Error sending request to bugzilla: %s", err)
 		return 0, err
 	}
+	if resp.StatusCode > 299 {
```
ameihm0912: Just curious why `299` here? Should we always just need `200`?

contrib/bugzilla-alert-manager/manager.go
ajvb: yep!

ameihm0912: 👍

PR #348, opened 2020-03-27
Rework state and batch Watchlist queries

ameihm0912:
This is a fairly substantial rework of the state API and how state functions.

State has been modified to support both standard queries and transactions. The underlying state interface is responsible for optimizing the query, which in the case of Datastore will result in use of the batch query APIs. Previously transactions were used across the board, this also modifies consumers so transactions are only used when required.

Update `PostProcessing` and `Watchlist` to make use of new batch query mode, and implement timers in watchlist analysis to handle periodic batching of watchlist lookups.


src/main/java/com/mozilla/secops/state/DatastoreStateCursor.java
```diff
@@ -1,81 +1,140 @@
...
+    }
+
+    Entity[] setParamFinal = setParam.toArray(new Entity[0]);
+    Key[] getParamFinal = getParam.toArray(new Key[0]);
+
+    Iterator<Entity> getResults = null;
+    List<Entity> setResults = null;
+    try {
+      getResults = rw.get(getParamFinal);
+      setResults = rw.put(setParamFinal);
```
ajvb: Is `setResults` supposed to be used anywhere?

ameihm0912:
> 🎉 looks awesome
> 
> Not needed imo, but would love to know the difference in performance for WatchlistAnalyze with this change.

@ajvb I don't have benchmarks from Dataflow runner but this is end to end for `testWatchlistAnalyze`:

Before:
```
TestPostProcessing.testWatchlistAnalyze: [measured 5 out of 7 rounds, threads: 1 (sequential)]
 round: 38.85 [+- 9.54], round.block: 0.00 [+- 0.00], round.gc: 0.00 [+- 0.00], GC.calls: 36, GC.time: 0.38, time.total: 238.51, time.warmup: 44.26, time.bench: 194.25
```

After:
```
TestPostProcessing.testWatchlistAnalyze: [measured 5 out of 7 rounds, threads: 1 (sequential)]
 round: 17.85 [+- 5.03], round.block: 0.00 [+- 0.00], round.gc: 0.00 [+- 0.00], GC.calls: 28, GC.time: 0.19, time.total: 110.55, time.warmup: 21.32, time.bench: 89.23
```

Round time includes the temporally bound tick generation, so I anticipate this should help with the batch lookups.

PR #353, opened 2020-04-01
Check for missing javadocs using checkstyle

ajvb:
Adds checkstyle:check within CircleCI building process that will check for missing javadoc comments from public methods. The configuration is stored in `checkstyle/*`

You can see an example of the CircleCI output for missing javadoc comments here: https://circleci.com/gh/mozilla-services/foxsec-pipeline/196

A few files have been excluded using `checkstyle/checkstyle-suppressions.xml` and a couple of in-use Beam Annotations (Setup, Teardown, ProcessElement, OnTimer, OnStale) have been excluded as well.

### DIFF
--- a/checkstyle/foxsec_checks.xml
+++ b/checkstyle/foxsec_checks.xml
@@ -36,8 +36,8 @@
         <module name="JavadocParagraph"/>
         <module name="JavadocMethod">
             <property name="scope" value="public"/>
-            <property name="allowMissingParamTags" value="true"/>
-            <property name="allowMissingReturnTag" value="true"/>
+            <property name="allowMissingParamTags" value="false"/>
+            <property name="allowMissingReturnTag" value="false"/>
             <property name="allowedAnnotations" value="Override, Test, Setup, Teardown, ProcessElement, OnTimer, OnStale"/>
             <property name="tokens" value="METHOD_DEF, CTOR_DEF, ANNOTATION_FIELD_DEF"/>
         </module>
--- a/checkstyle/foxsec_checks.xml
+++ b/checkstyle/foxsec_checks.xml
@@ -1,0 +1,51 @@
+<?xml version="1.0"?>
+<!DOCTYPE module PUBLIC
+          "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
+          "https://checkstyle.org/dtds/configuration_1_3.dtd">
+
+<!--
+    Checkstyle configuration specific to the foxsec-pipeline. This is lacking as we rely
+    heavily on spotless for the majority of linting.
+
+    Checkstyle is very configurable. Be sure to read the documentation at
+    http://checkstyle.org (or in your downloaded distribution).
+
+    To completely disable a check, just comment it out or delete it from the file.
+    To suppress certain violations please review suppression filters.
+ -->
+
+<module name = "Checker">
+    <property name="charset" value="UTF-8"/>
+
+    <property name="fileExtensions" value="java, properties, xml"/>
+    <!-- Excludes all 'module-info.java' files              -->
+    <!-- See https://checkstyle.org/config_filefilters.html -->
+    <module name="BeforeExecutionExclusionFileFilter">
+        <property name="fileNamePattern" value="module\-info\.java$"/>
+    </module>
+    <!-- https://checkstyle.org/config_filters.html#SuppressionFilter -->
+    <module name="SuppressionFilter">
+        <property name="file" value="${org.checkstyle.google.suppressionfilter.config}"
+                default="checkstyle/checkstyle-suppressions.xml" />
+        <property name="optional" value="true"/>
+    </module>
+
+    <module name="TreeWalker">
+        <module name="InvalidJavadocPosition"/>
+        <module name="JavadocTagContinuationIndentation"/>
+        <module name="JavadocParagraph"/>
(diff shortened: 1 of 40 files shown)


src/main/java/com/mozilla/secops/awsbehavior/AwsBehavior.java
```diff
@@ -79,6 +79,7 @@ public ParseAndWindow(AwsBehaviorOptions options) {
 
     private Logger log;
 
+    /** Initialize new Matcher with a {@link CloudtrailMatcher} */
```
ameihm0912, line 82: This is missing parameter documentation, should the check be capturing that too?

src/main/java/com/mozilla/secops/awsbehavior/AwsBehavior.java
```diff
@@ -147,6 +148,7 @@ public void processElement(ProcessContext c) {
     private CloudtrailMatcherManager cmmanager;
     private Logger log;
 
+    /** Initialize new Matchers with {@link AwsBehaviorOptions} */
```
ameihm0912, line 151: Same as above here

src/main/java/com/mozilla/secops/parser/FxaAuth.java
```diff
@@ -309,6 +309,11 @@ private Boolean discernPasswordForgotSendCode() {
     return false;
   }
 
+  /**
+   * Check if the auth event contained a successful certifcate signing
```
ameihm0912, line 313: Typo here

src/main/java/com/mozilla/secops/parser/KeyedEvent.java
```diff
@@ -15,13 +15,15 @@
   private final String key;
   private final Event event;
 
+  /** Convert KeyedEvent to {@link KV} */
```
ameihm0912, line 18: Missing return type

src/main/java/com/mozilla/secops/parser/KeyedEvent.java
```diff
@@ -15,13 +15,15 @@
...
   public KV<String, Event> toKV() {
     if ((key == null) || (event == null)) {
       return null;
     }
     return KV.of(key, event);
   }
 
+  /** Initialize new KeyedEvent */
```
ameihm0912, line 26: Missing parameters

src/main/java/com/mozilla/secops/postprocessing/PostProcessing.java
```diff
@@ -159,13 +159,15 @@ public int hashCode() {
       }
     }
 
+    /** Initialize WatchlistAnaylze with {@link PostProcessingOptions} */
```
ameihm0912, line 162: Missing parameters + typo

ajvb:
>Looks good but I am wondering if there is a way we can not need to copy all the inherited interface documentation and duplicate it. Right now the documentation for interface implementations gets copied from the interface itself. Is there some option we have to detect that so the check still passes?

So there doesn't seem to be a way to disable it without adding something to it. We could write our own noop annotation, but I found the javadoc tag `{@inheritDoc}` which does what we want. It also allows extending the inherited javadoc if we want to.

>Are these being suppressed here because they are just missing and we need to fix them?

For `UserIdentity.java` and `CloudtrailEvent.java`, yes. I added links to the AWS docs on them, but yeah, I feel like it would be wasted effort to copy all of the documentation for these into our model classes.

The `*Interface.java` files were there because I didn't know about the `{@inheritDoc}` tag. They have been removed.

The Kinesis files are there because they are throwing errors and don't think we should fix if they are going to get removed in a few weeks.

---

Additionally, I had originally disabled the "fail on missing return tag" and "fail on missing param tag", but I turned them back on and added missing tags in all of the comments.

checkstyle/foxsec_checks.xml
```diff
@@ -0,0 +1,51 @@
...
+    </module>
+
+    <module name="TreeWalker">
+        <module name="InvalidJavadocPosition"/>
+        <module name="JavadocTagContinuationIndentation"/>
+        <module name="JavadocParagraph"/>
+        <module name="JavadocMethod">
+            <property name="scope" value="public"/>
+            <property name="allowMissingParamTags" value="false"/>
+            <property name="allowMissingReturnTag" value="false"/>
```
ajvb, line 40: @ameihm0912 these two options allow for disabling missing param/return tags. Originally they were set to "true", changed to false now.

ameihm0912:
> So there doesn't seem to be a way to disable it without adding something to it. We could write our own noop annotation, but I found the javadoc tag `{@inheritDoc}` which does what we want. It also allows extending the inherited javadoc if we want to.

👍 looks good.

> For `UserIdentity.java` and `CloudtrailEvent.java`, yes. I added links to the AWS docs on them, but yeah, I feel like it would be wasted effort to copy all of the documentation for these into our model classes.

Makes sense yeah, I just wasn't sure if they needed to be fixed or if we were excluding them for a reason.

> The Kinesis files are there because they are throwing errors and don't think we should fix if they are going to get removed in a few weeks.

The fixes should be in Beam 2.21 so we can remove all that code then.

> Additionally, I had originally disabled the "fail on missing return tag" and "fail on missing param tag", but I turned them back on and added missing tags in all of the comments.

👍

PR #376, opened 2020-04-15
Option to include known gateways in nat detection logic.

### LINKED ISSUE #373: Support initialization of DetectNat with static address list
Currently `DetectNat` dynamically produces a side input as a materialized map based in ongoing analysis.

It should be modified so it can be optionally initialized with a static list of addresses, which will always be considered gateways by transforms that make use of NAT detection.

It would be good to just load this from GCS so a pattern like we use with [identity manager](https://github.com/mozilla-services/foxsec-pipeline/blob/dc192fc60579d607a4007de738709efe00b6c93e/src/main/java/com/mozilla/secops/identity/IdentityManager.java#L36-L40) could work, and the input option would be this path and the file would contain a list of addresses.

kkleemola:
closes #373


src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
```diff
@@ -1563,8 +1565,12 @@ public HTTPRequestAnalysis(HTTPRequestOptions options, HTTPRequestToggles toggle
...
     void setNatDetection(Boolean value);
 
-    @Description(
-        "Path to load user agent blacklist from for UA blacklist analysis; resource path, gcs path")
+    @Description("Path to load initial gateway list for nat detection; resource path, gcs path")
+    String getKnownGatewaysPath();
+
+    void setKnownGatewaysPath(String value);
+
+    @Description("Path to load initial gateway list for nat detection; resource path, gcs path")
```
ameihm0912, line 1573: Looks like incorrect description added here

src/main/java/com/mozilla/secops/NatUtil.java
```diff
@@ -0,0 +1,49 @@
...
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.HashMap;
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class NatUtil {
```
ameihm0912, line 12: Missing class documentation here

Should the function in here just be in `DetectNat` since it's only used there anyway?

src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
kkleemola: fixed, thanks.

src/main/java/com/mozilla/secops/DetectNat.java
```diff
@@ -49,57 +56,163 @@
...
+     * @param knownGateways map containing known gateway ips as keys and True as the value
+     * @return A {@code UserAgentBased} {@link PTransform}
+     */
+    @VisibleForTesting
+    public UserAgentBased withKnownGateways(Map<String, Boolean> knownGateways) {
+      return new UserAgentBased(knownGateways);
+    }
+  }
+
+  public static DetectNat.UserAgentBased byUserAgent() {
```
ajvb, line 181: nit: missing javadoc comment here.

(Something's up with the checkstyle javadoc check, looking into it atm)

src/main/java/com/mozilla/secops/DetectNat.java
```diff
@@ -49,57 +56,163 @@
...
+    return new UserAgentBased(new HashMap<String, Boolean>());
+  }
+
+  /**
+   * Load detect nat manager configuration from a resource file
+   *
+   * @param path Path to load file file from, resource path or GCS URL
+   * @return a map containing each ip and the value true
+   */
+  public static Map<String, Boolean> loadGatewayList(String path) {
```
ajvb, line 191: This looks great as is, but one quick thought. Should this be combined with the "named subnets" stuff in `IdentityManager` (https://github.com/mozilla-services/foxsec-pipeline/blob/master/src/main/java/com/mozilla/secops/identity/IdentityManager.java#L22) ? I feel like the purpose for these is overlapping a bit.

kkleemola: I think we do something similar also for user agent lists. I'll merge this but create a separate pr to clean up all the read from gcs or file path logic we have.

Issue #358, opened 2020-04-06
Make AlertMeta more strictly defined


ajvb:
Make [AlertMeta](https://github.com/mozilla-services/foxsec-pipeline/blob/master/src/main/java/com/mozilla/secops/alert/AlertMeta.java) more strictly defined rather than the current free form key strings everywhere.

Currently there are free form ("magic strings") everywhere for metedata key names. This is also present throughout the cloud functions (https://github.com/mozilla-services/foxsec-pipeline/blob/master/contrib/bugzilla-alert-manager/manager.go#L121).

We should make these keys more strictly defined and do this in a way that can be tested and checked for both the Java and Go code. 

Related to #182

ajvb:
```
$ ag getMetadataValue foxsec-pipeline/src/ | wc -l
604
```

PR #379, opened 2020-04-15
remove string usage in metadata

ameihm0912:
Related to #358

This removes all usage of strings as metadata keys in Java code. This
does not yet cover off the relationship with cloud functions, which
would be the next step.

What also needs to happen following this is a larger consolidation of
key names as there are a lot of duplicates and potentially unused
values, and we can probably reduce the total key count.

PRing this now because this diff is starting to become a bit unwieldy but this represents the major portion of what is required for #358 and we can piece out the rest in other PRs.


ameihm0912:
One additional note here, this PR also removes all the extra metadata being added in gatekeeper; this was largely not useful so we are eliminating it in favor of the using the source to get the details.

PR #387, opened 2020-04-22
Add metrics for nat detection.

kkleemola:


src/main/java/com/mozilla/secops/httprequest/HTTPRequestMetrics.java
```diff
@@ -0,0 +1,30 @@
...
+import org.apache.beam.sdk.metrics.Counter;
+import org.apache.beam.sdk.metrics.Metrics;
+
+/** {@link HTTPRequestMetrics} contains metrics for the {@link HTTPRequest} pipeline. */
+public class HTTPRequestMetrics {
+
+  /** Metrics for the various analysis transforms in {@link HTTPRequest} pipeline */
+  public static class HeuristicMetrics implements Serializable {
+    private static final long serialVersionUID = 1L;
+    private final String NAT_DETECTED = "nat_detected";
```
ameihm0912, line 13: nit: should this be something more like `alert_skipped_nat_detected` or similar since it applies to alerts we are suppressing and not actual detections? I'm fine leaving it as is either way.

PR #396, opened 2020-04-23
Mark fields sensitive and dynamically create metrics view

ameihm0912:
Extend metadata key map to make use of sensitive field flags.

Use this to generate metrics view, such that it can be automatically updated.


src/main/java/com/mozilla/secops/alert/AlertMeta.java
```diff
@@ -151,16 +142,16 @@ public boolean validate(String key, String value) {
...
-    USER_NAME("user_name"),
+    UID("uid", true),
     USERAGENT("useragent"),
     USERNAME("username"),
     URL_TO_FINDING("url_to_finding"),
     WINDOW_TIMESTAMP("window_timestamp"),
     WHITELISTED_ENTRY_CREATED_BY("whitelisted_entry_created_by"),
     SOURCEADDRESS(
         "sourceaddress",
+        true,
```
ajvb, line 153: Does this mark the associated keys as sensitive as well? I feel like some like city, country, and ISP would probably be sensitive as well?

ameihm0912: It doesn't mark the associated no, just the identifying field. The others are left as they can be used as aggregates in the metrics views.

PR #398, opened 2020-04-27
add heuristic for errors matching a specific endpoint pattern

### LINKED ISSUE #385: Path specific error rate analyzer
Create error rate analyzer similar to what exists currently in HTTPRequest, however can be:

- Configured to look for errors involving specific endpoints
- Does not wait for windows to be closed, providing early panes (potentially using session windows instead)

We already do session windows for one heuristic here, so it may be worth looking at if we can reuse the existing windowing path.

kkleemola:
This reuses the existing session windowing logic - so status is added to the "abridged" events that are used instead of the full ones.

The way this is implemented we take a pattern path and increment it for that.
So if our config was 3:GET:/test/(\\d+)/profile and we had requests for:
GET /test/1/profile 404
GET /test/2/profile 404
GET /test/3/profile 404

then we would have a count of 3 for the endpoint and generate an alert even though it's 3 different paths.

closes #385


src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
```diff
@@ -1451,21 +1635,33 @@ public HTTPRequestAnalysis(HTTPRequestOptions options, HTTPRequestToggles toggle
...
-                        new KeyAndWindowForSessionsFireEarly(toggles))
-                    // No requirement for follow up application of GlobalTriggers here since
-                    // EndpointAbuseAnalysis will do this for us
-                    .apply(
-                        "endpoint abuse analysis",
-                        new EndpointAbuseAnalysis(
-                            toggles,
-                            enableIprepdDatastoreWhitelist,
-                            iprepdDatastoreWhitelistProject)));
+      if (toggles.getEnableEndpointAbuseAnalysis()
```
ajvb, line 1638: Are there cases where someone could trigger both an alert from `EndpointAbuseAnalysis` and `PerEndpointErrorRateAnalysis` and get blocked too quickly? Or is it that if someone triggers both quick enough they're doing something nefarious enough to get blocked no matter?

src/test/java/com/mozilla/secops/httprequest/TestPerEndpointErrorRateAnalysis.java
```diff
@@ -0,0 +1,323 @@
...
+    PCollection<Long> count = results.apply(Count.globally());
+
+    PAssert.that(count).containsInAnyOrder(1L, 1L);
+
+    PAssert.that(results)
+        .satisfies(
+            i -> {
+              for (Alert a : i) {
+                if (a.getMetadataValue(AlertMeta.Key.SOURCEADDRESS).equals("192.168.1.2")) {
+                  assertEquals("1970-01-01T00:00:00.000Z", a.getTimestamp().toString());
```
ajvb, line 85: nit: should there be a check that these asserts have ran? or is the `PAssert.that(count).containsInAnyOrder(1L, 1L);` enough? Just wonder if there could be a case in the future where the if statement doesn't match and it's not caught?

src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
kkleemola: I don't think so? The endpoint abuse violation alone is enough to be blocked.

src/test/java/com/mozilla/secops/httprequest/TestPerEndpointErrorRateAnalysis.java
kkleemola: There should probably be a check. Added it.

kkleemola:
> This looks good overall, I had one question: should we take into account the time between the error requests that are being considered in the analysis step here? For example, since the session will only end after X minutes of inactivity, if we have a client that generates a few errors up front, makes valid requests for 30 minutes, then generates one more error should that result in a violation?

It does seem a bit off to do that, esp since the errors could be pretty accidental or not even all the same kind of error. Perhaps its better just to change this to use a sliding window instead of trying to create sliding windows on top of the session windows...

ameihm0912:
> It does seem a bit off to do that, esp since the errors could be pretty accidental or not even all the same kind of error. Perhaps its better just to change this to use a sliding window instead of trying to create sliding windows on top of the session windows...

Yea maybe a few diff options here?

* Leave as is, and add some sort of analysis of the time between the requests into the `DoFn`

Pro is we reuse the existing session windows, con is it's a bit inefficient and more complex.

* Filter 4xx and then sessionize with a shorter gap duration, reuse existing analysis logic.

We'd end up with another windowing path, but with this you could probably reuse your existing analysis logic and the windowed data since it only would contain error events would generally be pretty small.

* Other windowing options

The sliding windows is an option too; we've used these in the past but generally haven't much as with shorter "slides every" parameters the window density can start to become problematic, but if we limit this to just errors that might work OK too.

kkleemola:
I tried adding the timing analysis and using the existing session windowing but it becomes pretty complicated and messy. So I switched to filtering the non 4xx events and then creating session windows based on those (and there's a setting so it can use a shorter session gap duration).

I think this is a lot cleaner and easier to understand as well as will be more efficient.

src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
```diff
@@ -1623,6 +1893,29 @@ public HTTPRequestAnalysis(HTTPRequestOptions options, HTTPRequestToggles toggle
 
     void setEndpointSequenceAbusePatterns(String[] value);
 
+    @Description("Paths for per endpoint rate limit monitoring; e.g., threshold:method:path ")
```
ameihm0912, line 1896: `s/rate/error rate/`?

src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
```diff
@@ -1336,6 +1343,248 @@ private Boolean belongsToSequence(String method, String path) {
...
+      /** Path (pattern) */
+      public Pattern path;
+
+      /** Threshold */
+      public Integer threshold;
+
+      /** Interval */
+      public Long interval;
+
+      /** Returns true iff an event matches the endpoint for both method and path */
```
ameihm0912, line 1423: Small typo here

ameihm0912:
@kkleemola if possible can we rebase this off master and squash out the previous implementation from the history before we merge it?

kkleemola:
Fixed the typos + rebased & squashed out the old implementation.

Re:option hell, definitely agree. I think revisiting the config would be good to do soon as well as perhaps breaking up HttpRequest? (similar to how for customs each heuristic is separate if we don't mind losing some git history).

PR #405, opened 2020-05-19
Template with links for amo cloud submission alert

### LINKED ISSUE #404: Linkify text for add-on cloud submission slack message
It would be helpful to linkify some text for add-on cloud submission slack messages.

Current message style:
addons.mozilla.org cloud provider addon submission from 0.0.0.0, guid example@guid isapi true user_id 123456 (00000000-0000-0000-0000-000000000000)

Linkified message style:
addons.mozilla.org cloud provider addon submission from 0.0.0.0, guid [example@guid](https://addons-internal.prod.mozaws.net/en-US/admin/models/addons/addon/example@guid/change/) isapi true user_id [123456](https://addons-internal.prod.mozaws.net/en-US/admin/models/users/userprofile/123456/change/) (00000000-0000-0000-0000-000000000000)

Notes:
The link for the add-on GUID would be:
https://addons-internal.prod.mozaws.net/en-US/admin/models/addons/addon/%s/change/

The link for the user ID would be:
https://addons-internal.prod.mozaws.net/en-US/admin/models/users/userprofile/%s/change/

For Slack, the link format is:
`<https://addons-internal.prod.mozaws.net/en-US/admin/models/users/userprofile/123456/change/|123456>`

ajvb:
Adds support for "catchall" templates, exclusive to slack at the moment.
These will be specifically for templates that are meant to be sent to a
shared email or slack channel rather than individualized alerts (i.e. authprofile)

Fixes #404

### DIFF
--- a/src/main/java/com/mozilla/secops/alert/Alert.java
+++ b/src/main/java/com/mozilla/secops/alert/Alert.java
@@ -117,28 +117,6 @@ public class Alert implements Serializable {
   }
 
   /**
-   * Set a masked summary in the alert
-   *
-   * <p>Some output transforms will prefer the masked summary to the primary summary field, assuming
-   * the masked summary has sensitive information removed.
-   *
-   * @param maskedSummary Masked summary string
-   */
-  public void setMaskedSummary(String maskedSummary) {
-    addMetadata(AlertMeta.Key.MASKED_SUMMARY, maskedSummary);
-  }
-
-  /**
-   * Get any masked summary value in the alert
-   *
-   * @return Masked summary or null if unset
-   */
-  @JsonIgnore
-  public String getMaskedSummary() {
-    return getMetadataValue(AlertMeta.Key.MASKED_SUMMARY);
-  }
-
-  /**
    * Set alert merge key for notifications in metadata
    *
    * <p>If a merge key is set in metadata for an alert, some output transforms will utilize this key
@@ -445,6 +423,25 @@ public class Alert implements Serializable {
   @JsonIgnore
   public String getSlackTemplate() {
     return getMetadataValue(AlertMeta.Key.TEMPLATE_NAME_SLACK);
+  }
+
+  /**
+   * Set slack catchall template name
+   *
+   * @param templateName Freemarker template name with file extension
+   */
+  public void setSlackCatchallTemplate(String templateName) {
+    addMetadata(AlertMeta.Key.TEMPLATE_NAME_SLACK_CATCHALL, templateName);
+  }
+
+  /**
+   * Get slack catchall template name
+   *
+   * @return Freemarker template name with file extension or null if not set.
+   */
+  @JsonIgnore
+  public String getSlackCatchallTemplate() {
+    return getMetadataValue(AlertMeta.Key.TEMPLATE_NAME_SLACK_CATCHALL);
   }
 
   /**
--- a/src/main/java/com/mozilla/secops/amo/AddonCloudSubmission.java
+++ b/src/main/java/com/mozilla/secops/amo/AddonCloudSubmission.java
@@ -82,6 +82,7 @@ public class AddonCloudSubmission extends PTransform<PCollection<Event>, PCollec
                     alert.setCategory("amo");
                     alert.setSubcategory("amo_cloud_submission");
                     alert.setNotifyMergeKey("amo_cloud_submission");
+                    alert.setSlackCatchallTemplate(Amo.SLACK_CATCHALL_TEMPLATE);
                     alert.addMetadata(AlertMeta.Key.PROVIDER, f);
                     alert.addMetadata(AlertMeta.Key.SOURCEADDRESS, d.getRemoteIp());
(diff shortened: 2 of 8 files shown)


src/main/java/com/mozilla/secops/alert/Alert.java
```diff
@@ -116,28 +116,6 @@ public String getSummary() {
...
 
-  /**
-   * Set a masked summary in the alert
-   *
-   * <p>Some output transforms will prefer the masked summary to the primary summary field, assuming
-   * the masked summary has sensitive information removed.
-   *
-   * @param maskedSummary Masked summary string
-   */
-  public void setMaskedSummary(String maskedSummary) {
```
ajvb, line 127: @ameihm0912 I didn't see this being called from anywhere, so I removed this. Let me know if we should keep it though.

src/main/java/com/mozilla/secops/CompositeOutput.java
```diff
@@ -47,6 +47,11 @@ private CompositeOutput() {}
         alertcfg.registerTemplate(tmpl);
       }
     }
+    if (options.getOutputAlertCatchallTemplates() != null) {
```
ameihm0912, line 50: Is this option needed? Could AMO just register the templates it knows it needs without the option?

src/main/java/com/mozilla/secops/alert/Alert.java
ameihm0912: Sounds good, I think we were using this prior but can just remove it now.

src/main/java/com/mozilla/secops/amo/AddonCloudSubmission.java
```diff
@@ -82,6 +82,8 @@ public void processElement(ProcessContext c) {
                     alert.setCategory("amo");
                     alert.setSubcategory("amo_cloud_submission");
                     alert.setNotifyMergeKey("amo_cloud_submission");
+                    alert.setSlackCatchallTemplate(Amo.SLACK_CATCHALL_TEMPLATE);
+                    alert.addMetadata(AlertMeta.Key.MONITORED_RESOURCE, monitoredResource);
```
ameihm0912, line 86: Was adding this needed for the template? It would usually get added in the alerting output path if it's missing, is this here because we don't call the alert formatter in the tests?

src/main/java/com/mozilla/secops/CompositeOutput.java
ajvb: I'm open to either, but I do think that it would be worth separating out the "for a specific individual" templates from the "for a catchall email/slack channel" in regards to the options and registering of the templates. If we want to add a catchall template for authprofile, it will be very clear with this implementation which templates are for which "alert channels".

src/main/java/com/mozilla/secops/amo/AddonCloudSubmission.java
ajvb: Aw yeah, thanks, fixing.

src/main/java/com/mozilla/secops/CompositeOutput.java
ameihm0912: I'm not too sure what the advantage would be. All the template registration does is inform the template manager which ones to load off GCS or from jar resources when it starts.

You still wouldn't be able to tell which are used where unless you looked at the individual classes.

src/test/java/com/mozilla/secops/amo/TestAmo.java
```diff
@@ -214,6 +216,32 @@ public void amoFxaAbuseNewVersionTest() throws Exception {
                         "test cloud provider addon submission from 52.204.100.1, guid extension_guid"
                             + " isapi true user_id 99999999",
                         a.getSummary());
+                    assertEquals("slack/catchall/amo.ftlh", a.getSlackCatchallTemplate());
+
+                    // Test slack template rendereing
+                    try {
+                      // Add it here, where it usually would get added in the Alert formatter
+                      a.addMetadata(AlertMeta.Key.MONITORED_RESOURCE, "test");
```
ameihm0912, line 224: You can probably get away from needing to add this if you slightly modify the code above the `PAssert`.

`AlertFormatter` is actually called which will add the resource field, but it's called as part of executing the composite output step which the assert doesn't look at. You could probably adjust things a bit there, e.g.,

```java
PCollection<Alert> alerts = Amo.executePipeline(p, input, options)
    .apply(ParDo.of(new AlertFormatter(getTestOptions())));
```

src/main/java/com/mozilla/secops/CompositeOutput.java
ajvb: Right right, I was getting the registration conflated with the template assignment. Fixed.

src/test/java/com/mozilla/secops/amo/TestAmo.java
ajvb: Ah ty. Fixed.

PR #407, opened 2020-05-20
Use radix tree for CIDR lookup in CidrUtil

ameihm0912:
CidrUtil is currently very inefficient in the sense for each check, it walks a flat list from beginning to end to see if an IP address matches any configured subnet. This issue manifests itself the most when combined with the parser, where we apply a check for every single event.

Convert inet4 lookups to make use of a radix tree which increases the performance of these methods by orders of magnitude.

### DIFF
--- a/src/main/java/com/mozilla/secops/InetRadix.java
+++ b/src/main/java/com/mozilla/secops/InetRadix.java
@@ -1,50 +1,48 @@
 package com.mozilla.secops;
 
-import java.net.InetAddress;
-import java.net.UnknownHostException;
+import com.google.common.net.InetAddresses;
 import java.nio.ByteBuffer;
 
 /**
  * CIDR lookup using radix tree search
  *
  * <p>Only supports IPv4.
+ *
+ * <p>See also http://www.cs.columbia.edu/~ji/F02/ir04/routing.pdf
  */
 public class InetRadix {
   private static class Node {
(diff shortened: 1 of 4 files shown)


src/main/java/com/mozilla/secops/InetRadix.java
```diff
@@ -0,0 +1,137 @@
...
+          next = n.l;
+        }
+        if (next == null) {
+          break;
+        }
+        bit = bit >> 1;
+        n = next;
+      }
+
+      if (next != null) {
```
kkleemola, line 69: mentioned above, but maybe replace the node so we prune the tree.
i.e. if we had a.b.c.d/32 and then a.b.c.d/24 added, we can get rid of some of the excess nodes and keep the minimum elements in the tree?

src/main/java/com/mozilla/secops/InetRadix.java
```diff
@@ -0,0 +1,137 @@
...
+      return ret;
+    }
+
+    public void insert(long ip, long mask) {
+      long bit = startBit;
+
+      Node n = root;
+      Node next = n;
+
+      while ((bit & mask) != 0) {
```
kkleemola, line 56: I think if this was also for loops over the ip, you could even skip creating a mask and just take the integer from the cidr. (i.e. just take the /24)

PR #415, opened 2020-05-29
Beam 2.21.0

ameihm0912:
* Upgrade to Beam 2.21.0
* Remove the temporary Kinesis code that was needed to resolve an issue that is now fixed in Beam


ameihm0912:
> might want to remove these lines as well: https://github.com/mozilla-services/foxsec-pipeline/blob/master/checkstyle/checkstyle-suppressions.xml#L8-L15

Good catch, will do that before merge.

PR #418, opened 2020-06-01
add monitor only velocity heuristic

kkleemola:


src/main/java/com/mozilla/secops/customs/CustomsVelocity.java
```diff
@@ -263,6 +275,68 @@ public void processElement(ProcessContext c) {
...
+                                      geoRespMO.getKmDistance(),
+                                      geoRespMO.getTimeDifference()));
+
+                              // It's possible the AlertFormatter DoFn could add this for us
+                              // later, but
+                              // since it is important information as part of this transform make
+                              // sure
+                              // it will be present by leveraging the formatters GeoIP method
+                              // here.
+                              AlertFormatter.addGeoIPData(alertMO, geoip);
```
ajvb, line 334: This can get removed if we want, as AlertFormatter is now doing this, but imo can also be done as a follow up.

src/main/java/com/mozilla/secops/customs/CustomsVelocity.java
```diff
@@ -263,6 +275,68 @@ public void processElement(ProcessContext c) {
...
                         }
+
+                        // for monitoring smaller velocity jumps than we escalate for
+                        if (checkExperimentalParam) {
+                          AuthStateModel.GeoVelocityResponse geoRespMO =
+                              sm.geoVelocityAnalyzeLatest(maxKilometersPerSecondMonitorOnly);
+
+                          if (geoRespMO != null) {
+                            log.info(
+                                "{}: new location is {}km away from last location within {}s",
```
ameihm0912, line 287: Should we have something in here to differentiate it from the other one?

src/main/java/com/mozilla/secops/customs/CustomsVelocity.java
```diff
@@ -263,6 +275,68 @@ public void processElement(ProcessContext c) {
...
+                                geoRespMO.getKmDistance(),
+                                geoRespMO.getTimeDifference());
+
+                            boolean minDistanceMetMO = true;
+                            if (minimumDistanceForAlertMonitorOnly != null) {
+                              if (geoRespMO.getKmDistance() < minimumDistanceForAlertMonitorOnly) {
+                                log.info(
+                                    "{}: will skip alert as minimum distance was not met (monitor only)",
+                                    uid);
+                                minDistanceMet = false;
```
ameihm0912, line 298: Should this be `minDistanceMetMO`?

src/main/java/com/mozilla/secops/customs/CustomsVelocity.java
```diff
@@ -263,6 +276,60 @@ public void processElement(ProcessContext c) {
...
+                                  AlertMeta.Key.TIME_DELTA_SECONDS,
+                                  geoRespMO.getTimeDifference().toString());
+                              alertMO.addMetadata(
+                                  AlertMeta.Key.KM_DISTANCE,
+                                  String.format("%.2f", geoRespMO.getKmDistance()));
+                              alertMO.addMetadata(AlertMeta.Key.UID, uid);
+                              alertMO.addMetadata(AlertMeta.Key.EMAIL, email);
+                              alertMO.setSummary(
+                                  String.format(
+                                      "%s %s velocity exceeded, %.2f km in %d seconds",
```
ameihm0912, line 323: We should probably have something in here to differentiate it as well

src/main/java/com/mozilla/secops/customs/CustomsVelocity.java
kkleemola: added.

src/main/java/com/mozilla/secops/customs/CustomsVelocity.java
kkleemola: yes. fixed + added testcase for it.

src/main/java/com/mozilla/secops/customs/CustomsVelocity.java
kkleemola: added.

src/main/java/com/mozilla/secops/customs/CustomsVelocity.java
kkleemola: Will do separately as we can do it for the non monitor alert too and it requires changing tests.

PR #428, opened 2020-06-17
exclude 429 from per endpoint error rate analysis and escalate to iprepd

### LINKED ISSUE #427: Modify PerEndpointErrorRateAnalysis to respond more quickly and use a higher violation penalty
Modify `PerEndpointErrorRateAnalysis` to:

- Emit alerts more quickly, if possible
- Be convertible into a higher impact violation (e.g., 75)

kkleemola:
1) Adds escalation to iprepd using the generic violations.
2) Excludes 429s from the error session as these requests were already blocked.

fixes #427

### DIFF
--- a/docker/iprepd.yaml
+++ b/docker/iprepd.yaml
@@ -24,6 +24,27 @@ violations:
     penalty: 100
   - name: hard_limit_violation
     penalty: 10
+  - name: violation5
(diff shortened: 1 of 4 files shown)


kkleemola:
Updated this so we don't exclude 429s after all.

Issue #432, opened 2020-06-18
Remove exclusionary language


kkleemola:
Two areas I found in which we use "whitelist" / "blacklist" instead of alternatives:

1) User Agent "blacklist" heuristic
For this the heuristic name is written to the alerts db so we either need to remap the historical data to the new name or be okay with our dashboards having a mix of both for some time.

2) Iprepd "whitelist"

For this, we need to change the datastore namespace as well but since there's so few objects (3) in it its a relatively easy migration.

We also need to change the slackbot code so we rename the slash command as well.

PR #436, opened 2020-06-25
rename iprepd whitelist to exempt

kkleemola:
The other half of #432 

Note: the slash commands for FoxSecBot have changed so we'll need to update them.

### DIFF
--- a/src/main/java/com/mozilla/secops/postprocessing/PostProcessing.java
+++ b/src/main/java/com/mozilla/secops/postprocessing/PostProcessing.java
@@ -361,7 +361,7 @@ public class PostProcessing implements Serializable {
       a.addMetadata(AlertMeta.Key.MATCHED_METADATA_VALUE, k.value);
       a.addMetadata(AlertMeta.Key.MATCHED_TYPE, entry.getType());
       a.addMetadata(AlertMeta.Key.MATCHED_OBJECT, entry.getObject());
-      a.addMetadata(AlertMeta.Key.EXEMPTION_CREATED_BY, entry.getCreatedBy());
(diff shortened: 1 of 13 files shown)


ajvb:
@kkleemola What is the migration process/steps for this?

src/main/java/com/mozilla/secops/postprocessing/PostProcessing.java
```diff
@@ -361,7 +361,7 @@ private Alert createAlert(Alert sourceAlert, Watchlist.WatchlistEntry entry, Key
       a.addMetadata(AlertMeta.Key.MATCHED_METADATA_VALUE, k.value);
       a.addMetadata(AlertMeta.Key.MATCHED_TYPE, entry.getType());
       a.addMetadata(AlertMeta.Key.MATCHED_OBJECT, entry.getObject());
-      a.addMetadata(AlertMeta.Key.WHITELISTED_ENTRY_CREATED_BY, entry.getCreatedBy());
+      a.addMetadata(AlertMeta.Key.EXEMPTION_CREATED_BY, entry.getCreatedBy());
```
ajvb, line 364: Oh hm. I believe this was a bug. This should be `WATCHLIST_CREATED_BY`

kkleemola:
I believe the only other migration step is copying the datastore entries to the new namespace. But because there's only three items in there I was just going to manually add those ips with the slack command. (We could also write a small script to copy datastore entries to a different namespace, but not sure it's worth it for 3 items)

src/main/java/com/mozilla/secops/postprocessing/PostProcessing.java
kkleemola: Nice catch. Looks like that was only used by watchlist so I removed exemption and added watchlist.

PR #448, opened 2020-07-08
enable CHMCache in maxmind reader

ameihm0912:


ajvb:
Would we want to increase the capacity beyond the [default](https://github.com/maxmind/MaxMind-DB-Reader-java/blob/master/src/main/java/com/maxmind/db/CHMCache.java#L15)?

ameihm0912:
Closing this for now (see #450)

PR #456, opened 2020-07-17
Initial Pioneer analysis pipeline

ameihm0912:
Composite transform for data exfiltration analysis specific to Pioneer configuration.

There are some TODO's here still, related to actual alerting metadata and routing which will be described in more detail in the associated bug.


src/main/java/com/mozilla/secops/alert/AlertMeta.java
```diff
@@ -146,6 +147,7 @@ public boolean validate(Key key, String value) {
     IDENTITY_KEY("identity_key"),
     IDENTITY_UNTRACKED("identity_untracked"),
     INDICATOR("indicator"),
+    INSTANCENAME("instancename"),
```
kkleemola, line 150: nit: instance_name ?

src/main/java/com/mozilla/secops/pioneer/Pioneer.java
```diff
@@ -0,0 +1,306 @@
...
+  public static class PioneerExfiltration extends PTransform<PCollection<Event>, PCollection<Alert>>
+      implements DocumentingTransform {
+    private static final long serialVersionUID = 1L;
+
+    private final int thresholdBytes;
+    private final int thresholdMillis;
+    private final String monitoredResource;
+
+    // The gap duration for us to consider a flow session as expired
+    private static final Duration SESSION_GAP_DURATION = Duration.standardMinutes(30);
```
kkleemola, line 68: Should this be configurable or based on exfil threshold? Or as an alt, should we have supported ranges for exfil period documented/enforced somewhere?

src/main/java/com/mozilla/secops/alert/AlertMeta.java
ameihm0912: Agree, will fix

src/main/java/com/mozilla/secops/pioneer/Pioneer.java
ameihm0912: Yeah, ideally we could just derive this value from the configured time based threshold but I am not sure on the correlation. This is just a first pass so we can begin testing, so I think once we run through a few scenarios we can probably come up with a better way to deal with this.

ameihm0912:
> LGTM
> 
> Not needed at this point, but will definitely want to document how to investigate these alerts.

Yeah for sure, that part we need to sort out still I'm not quite sure how these will be routed.

PR #459, opened 2020-07-22
do not replace original message upon receiving operator response

### LINKED ISSUE #446: Include "Fraud User Guide" in slack confirmation message
If a user clicks the "This was me" button but then has further questions they will need to have remembered the link, as the user guide link is removed after clicking the button

Would be in https://github.com/mozilla-services/foxsec-pipeline/blob/master/contrib/slackbot-background/handlers.go#L98-L143

kkleemola:
Fixes #446 

This changes how the Slackbot response is set so it no longer replaces the original message that has the alert details including the operator documentation. Instead the original message remains and the message thanking the operator is sent as a new message from slackbot.

It does mean the action items (boxes for yes and no) are still present but we do have a flow for if they're clicked and an alert is not new.

Alternative would be to insert just info we want to message or if there's multiple fields we want to preserve (which seems like it may be the case), we could retemplate the message with the alert metadata.

### DIFF
--- a/contrib/slackbot-background/handlers.go
+++ b/contrib/slackbot-background/handlers.go
@@ -106,7 +106,7 @@ func handleAlertConfirm(ctx context.Context, callback common.InteractionData, db
 
 	response := &slack.Msg{
 		Text:            "Error responding; please contact SecOps (secops@mozilla.com)",
-		ReplaceOriginal: true,
+		ReplaceOriginal: false,
 	}
 
 	if !alert.IsStatus(common.ALERT_NEW) {


kkleemola:
I can try just adding the doc link and keeping a replacement message. Only thing was I had gotten a complaint about the info being gone once the alert is actioned and so thought was it might be less confusing to preserve the text. We can try just the doc_link tho as I suspect no matter what we won't be able to please everyone.

ameihm0912:
> I can try just adding the doc link and keeping a replacement message. Only thing was I had gotten a complaint about the info being gone once the alert is actioned and so thought was it might be less confusing to preserve the text. We can try just the doc_link tho as I suspect no matter what we won't be able to please everyone.

Ok sounds good yeah, we can just leave as is and see how that works.

PR #465, opened 2020-07-31
private relay analysis transform

### LINKED ISSUE #444: Implement Private Relay comparison algorithm
Implementation of comparison algorithm for Private Relay forward notifications and FxA RP callbacks

See also https://bugzilla.mozilla.org/show_bug.cgi?id=1626333

ameihm0912:
Implement analysis transform for private relay based on https://bugzilla.mozilla.org/show_bug.cgi?id=1626333.

This is being added to the customs pipeline, since I don't think this needs a dedicated pipeline and this will make it easier to correlate this data directly with FxA if we need to do so in the future.

/cc @groovecoder

Closes #444


src/main/java/com/mozilla/secops/customs/CustomsPreFilter.java
```diff
@@ -5,52 +5,63 @@
...
-import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.TupleTag;
 
 /**
  * Basic filtering of ingested events prior to analysis application
  *
- * <p>This transform will filter any events from the input collection that are not returned in
- * {@link Customs#featureSummaryRegistration}.
+ * <p>This DoFn will split events up based on type. Configuration ticks are emitted in the main
+ * output collection with contains FxA authentication server events.
```
ajvb, line 14: typo: which instead of with?

src/main/java/com/mozilla/secops/customs/PrivateRelayForward.java
```diff
@@ -0,0 +1,264 @@
...
+                      PrivateRelayForwardState prfs;
+                      try {
+                        prfs = curs.get(uid);
+                      } catch (StateException exc) {
+                        log.error("error fetching state for {}: {}", uid, exc.getMessage());
+                        return;
+                      }
+                      if (prfs == null) {
+                        // No state for the UID was found, so initialize a new state instance but
+                        // don't set the real address value which will indicate it is new
```
ajvb, line 189: The logic in this function lgtm, but this comment is kind of confusing: "...don't set the real address value which will indicate it is new". Where do we consider whether it is "new"? What does that mean? Isn't it just that we will set the address later in the checks? Is this saying "don't set the real address yet so that we know that it is new?"

Maybe I'm spacing on a Friday, but yeah this part of the comment confused me.

src/main/java/com/mozilla/secops/customs/CustomsPreFilter.java
ameihm0912: Will fix this up 👍

src/main/java/com/mozilla/secops/customs/PrivateRelayForward.java
ameihm0912: Good call here too, I actually rewrote this comment twice because it wasn't great before either. I'll see if we can fix this up a bit, the idea is if that value is `null` we assert it's a new entry but maybe it would be better if we had some sort of boolean flag to indicate that. I'll address this as well before we merge.

PR #475, opened 2020-09-08
Replace filter step w/ output tuple

kkleemola:
Replaces filtering the entire input for each resource with a single filter step with a pcollection tuple where each resource has a tag with id=resource indicator.


ameihm0912 (approved):
Looks good 👍 just noticed one small thing

ajvb (approved):

PR #477, opened 2020-09-10
add gbk+ungroup to avoid fusion in hard limit analysis

kkleemola:
follow up to #476 

Dataflow fuses the steps together so adding an extra step doesn't actually help unless we also add something to prevent fusion. There's a few options to prevent this: https://cloud.google.com/dataflow/docs/guides/deploying-a-pipeline#fusion-optimization

Adding a reshuffle step would be a slightly nicer looking alternative but it is marked as deprecated by beam.

### DIFF
--- a/src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
+++ b/src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
@@ -26,7 +26,6 @@ import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
@@ -346,6 +345,7 @@ public class HTTPRequest implements Serializable {
     }
 
     @Override
+    @SuppressWarnings("deprecation")
     public PCollection<Alert> expand(PCollection<Event> input) {
       if (natView == null) {
         // If natView was not set then we just create an empty view for use as the side input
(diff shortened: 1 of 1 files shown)


src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
```diff
@@ -366,8 +367,22 @@ public void processElement(ProcessContext c) {
                       c.output(n.getSourceAddress());
                     }
                   }))
-          .apply(Count.<String>perElement())
-          .apply(Filter.by(new HasCountGreaterThan()))
+          .apply("hard limit per client count", Count.<String>perElement())
+          .apply("filter clients above limit", Filter.by(new HasCountGreaterThan()))
+          .apply("force materialization of input", GroupByKey.create())
+          .apply(
+              "ungroup keys",
```
ameihm0912, line 374: We do currently use `Reshuffle` now already so using it again wouldn't be so bad:

https://github.com/mozilla-services/foxsec-pipeline/blob/master/src/main/java/com/mozilla/secops/gatekeeper/GuardDutyTransforms.java#L305-L308

if we prefer to try that to make the code more readable we could, this works too though if we don't want to suppress that deprecation notice.

ameihm0912 (approved):

ajvb (approved):

src/main/java/com/mozilla/secops/httprequest/HTTPRequest.java
kkleemola: I'll switch to the reshuffle then if we're already using it else where.

PR #482, opened 2020-09-25
Customs account enumeration detection

kkleemola:


src/main/java/com/mozilla/secops/customs/CustomsAccountEnumeration.java
```diff
@@ -0,0 +1,152 @@
...
+            "status check analysis",
+            ParDo.of(
+                    new DoFn<KV<String, CustomsFeatures>, KV<String, Alert>>() {
+                      private static final long serialVersionUID = 1L;
+
+                      @ProcessElement
+                      public void processElement(ProcessContext c) {
+                        String ipAddr = c.element().getKey();
+                        CustomsFeatures cf = c.element().getValue();
+                        // If the total status checks is less than threshold we
```
ajvb, line 80:
```suggestion
                        // If the total status check count is less than the threshold we
```

src/test/java/com/mozilla/secops/customs/TestCustoms.java
```diff
@@ -754,4 +752,241 @@ public void privateRelayForwardTest() throws Exception {
...
+    options.setContentServerVarianceMinClients(3L);
+
+    PCollection<Alert> alerts = Customs.executePipeline(p, p.apply(s), options);
+
+    PAssert.that(alerts)
+        .satisfies(
+            x -> {
+              int cnt = 0;
+              for (Alert a : x) {
+                if (a.getMetadataValue(AlertMeta.Key.SOURCEADDRESS).equals("10.0.0.1")) {
```
ajvb, line 865: Might want to keep `ip1Cnt` here as well. I'm sure this test is working fine, but I had a case with the gatekeeper stuff where it started failing (alert was generated, but with broken values) and using this exact pattern I missed it for a while.

src/test/java/com/mozilla/secops/customs/TestCustoms.java
```diff
@@ -754,4 +752,241 @@ public void privateRelayForwardTest() throws Exception {
...
```
ajvb, line 940: Same here with the `ip1Cnt`

src/test/java/com/mozilla/secops/customs/TestCustoms.java
```diff
@@ -754,4 +752,241 @@ public void privateRelayForwardTest() throws Exception {
...
+    /**
+     * Test set for a second window contains multiple ips: 10.0.0.1 crosses the threshold with a mix
+     * of successful and blocked requests 10.0.0.5 does not cross the threshold 10.0.0.6 crosses the
+     * threshold with only successful requests 10.0.0.7 crosses the request count threshold but
+     * doesn't have enough distinct accounts checked to trigger an alert
+     */
+    String[] eb3 = TestUtil.getTestInputArray("/testdata/customs_account_status_abuse2.txt");
+    /**
+     * Test set containing content server events within the second window Contains events for
+     * 10.0.0.1 and 10.0.0.6.
```
ajvb, line 911: If there are content server events for 10.0.0.1 then why is there still an alert generated?

src/main/java/com/mozilla/secops/parser/FxaContent.java
```diff
@@ -0,0 +1,135 @@
...
+    SIGNUP,
+    VALIDATE_EMAIL_DOMAIN,
+    OTHER
+  }
+
+  private void discernRequestType() {
+    String path = fxaContentData.getPath();
+
+    if (path != null) {
+      if (path.startsWith("/metrics-flow?")) {
```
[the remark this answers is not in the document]
kkleemola: That's a good point! Changed.

src/main/java/com/mozilla/secops/customs/ContentServerVarianceDetector.java
```diff
@@ -0,0 +1,94 @@
...
+import org.apache.beam.sdk.transforms.windowing.AfterWatermark;
+import org.apache.beam.sdk.transforms.windowing.FixedWindows;
+import org.apache.beam.sdk.transforms.windowing.Window;
+import org.apache.beam.sdk.values.KV;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PCollectionView;
+import org.apache.beam.sdk.values.TypeDescriptors;
+import org.joda.time.Duration;
+
+public class ContentServerVarianceDetector {
```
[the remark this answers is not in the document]
kkleemola: The idea was that eventually we might want to tune and require certain content server requests (i.e. only metrics etc) but for now I thought it was easier and we'll be less prone to FP if we just consider accessing any content server resources as enough (and following the naming conventions of variance substrings from EndpointAbuse in HttpRequest). But I agree it is rather confusing and not necessarily a good name. I was just struggling a bit to come up with a better one
I had renamed it to ContentServerResourceAccess and then didn't like that either, but I can do that or if you have any other suggestions.

src/test/java/com/mozilla/secops/customs/TestCustoms.java
kkleemola: Added, thank you!

src/test/java/com/mozilla/secops/customs/TestCustoms.java
kkleemola: The test has two windows: the alert is generated for 10.0.0.1 in window 1 and no alert should be generated in window 2. (Idea was to make sure side input windows get aligned).

But uh, then I didn't include assertions for the timestamp or emails checked. :( Added that.

src/main/java/com/mozilla/secops/customs/ContentServerVarianceDetector.java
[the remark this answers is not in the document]
ajvb: That reasoning makes sense. It might be worth including some of that information in a class comment?

src/main/java/com/mozilla/secops/parser/FxaContent.java
```diff
@@ -0,0 +1,135 @@
+package com.mozilla.secops.parser;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.io.IOException;
+import java.util.ArrayList;
+
+public class FxaContent extends SourcePayloadBase {
```
ajvb, line 8: Missing a javadoc class comment here

As a note, I don't think that our Javadoc check in circle (https://github.com/mozilla-services/foxsec-pipeline/blob/master/.circleci/config.yml#L15-L16) is working as expected :(

src/main/java/com/mozilla/secops/customs/ContentServerVarianceDetector.java
```diff
@@ -0,0 +1,103 @@
...
+   * <p>This is currently meant to work with the fixed 10 minute window based heuristics.
+   *
+   * @param events Input events
+   * @return {@link PCollectionView} representing output of analysis
+   */
+  public static PCollectionView<Map<String, Boolean>> getView(PCollection<Event> events) {
+    return events
+        .apply(
+            "fixed ten min variance window",
+            Window.<Event>into(FixedWindows.of(Duration.standardMinutes(10)))
```
ameihm0912, line 62: Does this need to align with which windowing algorithm is being used in the transform consuming this side input? If so I was wondering if we should create a constant for the 10 minute period or something so that we don't end up in a case where the windowing is modified but the side input isn't changed?

src/main/java/com/mozilla/secops/customs/CustomsAccountEnumeration.java
```diff
@@ -0,0 +1,152 @@
...
+
+                        ArrayList<String> accts = new ArrayList<>();
+                        events.forEach(e -> accts.add(CustomsUtil.authGetEmail(e)));
+                        long distinctAccts =
+                            accts.stream().distinct().filter(Objects::nonNull).count();
+                        if (distinctAccts < threshold) {
+                          return;
+                        }
+
+                        if (cf.nominalVariance()) {
```
ameihm0912, line 101: Do we want something in here to determine if anything _other_ then `status` was used, and if so return?

src/main/java/com/mozilla/secops/parser/FxaContent.java
```diff
@@ -0,0 +1,135 @@
...
+  private static final long serialVersionUID = 1L;
+
+  private com.mozilla.secops.parser.models.fxacontent.FxaContent fxaContentData;
+  private RequestType requestType;
+
+  public enum RequestType {
```
ameihm0912, line 14: missing javadoc here

src/main/java/com/mozilla/secops/parser/FxaContent.java
```diff
@@ -0,0 +1,135 @@
...
+   * Fetch parsed FxA content data
+   *
+   * @return FxA content data
+   */
+  @JsonProperty("fxacontent_data")
+  public com.mozilla.secops.parser.models.fxacontent.FxaContent getFxaContentData() {
+    return fxaContentData;
+  }
+
+  public RequestType getRequestType() {
```
ameihm0912, line 79: missing comment here

ameihm0912 (approved):

src/main/java/com/mozilla/secops/customs/ContentServerVarianceDetector.java
kkleemola: Yep! That's a good point. I added a helper for creating the ten minute windows.

src/main/java/com/mozilla/secops/customs/CustomsAccountEnumeration.java
kkleemola: Yep, that makes sense. Added.

ajvb (approved):

ameihm0912 (approved):

PR #490, opened 2020-10-27
Update Taskcluster's log model

ajvb:
per changes in https://github.com/taskcluster/taskcluster/pull/3615

additionally, adds `apiVersion` to the model in case we need to support multiple versions in the future.

### DIFF
--- a/src/test/java/com/mozilla/secops/parser/ParserTest.java
+++ b/src/test/java/com/mozilla/secops/parser/ParserTest.java
@@ -930,7 +930,23 @@ public class ParserTest {
 
   @Test
   public void testParseTaskcluster() throws Exception {
-    String buf =
+    String bufV1 =
+        "{\"insertId\":\"AAAAAAAAAAAAAAA\",\"jsonPayload\":{\"EnvVersion\":\"2.0\",\"Fields"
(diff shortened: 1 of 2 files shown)


src/test/java/com/mozilla/secops/parser/ParserTest.java
```diff
@@ -933,8 +933,8 @@ public void testParseTaskcluster() throws Exception {
     String buf =
         "{\"insertId\":\"AAAAAAAAAAAAAAA\",\"jsonPayload\":{\"EnvVersion\":\"2.0\",\"Fields"
             + "\":{\"apiVersion\":\"v1\",\"clientId\":\"mozilla-auth0/ad|Mozilla-LDAP|riker/servi"
```
kkleemola, line 935: nit: Can apiVersion be bumped to v2 so the testcase is a valid log example?

kkleemola (approved):

ameihm0912 (commented):
I'm wondering if this change should handle both formats for compatibility?

ajvb:
@ameihm0912 yeah agreed. Added `hasAuthed` as an alias to the new `getAuthenticated()` and updated the test.

ameihm0912 (approved):

PR #498, opened 2020-11-11
always include supplementary slack notification for crit objects

kkleemola:
As requested in bugzilla 1676426.
Always include the slack notification for critical objects. (i.e. so in hours access is slack notification only, out of hours is email + slack)

### DIFF
--- a/src/main/java/com/mozilla/secops/authprofile/CritObjectAnalyze.java
+++ b/src/main/java/com/mozilla/secops/authprofile/CritObjectAnalyze.java
@@ -1,5 +1,7 @@
 package com.mozilla.secops.authprofile;
 
+import static java.util.Optional.ofNullable;
+
 import com.mozilla.secops.DocumentingTransform;
 import com.mozilla.secops.alert.Alert;
 import com.mozilla.secops.alert.AlertMeta;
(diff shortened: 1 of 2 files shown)


src/main/java/com/mozilla/secops/authprofile/CritObjectAnalyze.java
```diff
@@ -93,6 +93,18 @@ public String getTransformDoc() {
 
   private void addEscalationMetadata(Alert a) {
     if (alternateCritSlackEscalation != null) {
+      // always add the slack escalation metadata if we have slack escalation enabled
+      // even if it's not within the policy
+      a.addMetadata(AlertMeta.Key.NOTIFY_SLACK_SUPPLEMENTARY, altEscalateChannel);
+      a.addMetadata(
+          AlertMeta.Key.SLACK_SUPPLEMENTARY_MESSAGE,
+          String.format(
+              "<!channel> critical authentication event observed %s to %s, %s [%s/%s]",
```
ajvb, line 102: Do we really want to after-hours `@channel` each time along side the ping? Could the message also be slightly different to include that a page was sent as well?

kkleemola: That's fair. Changed it a bit.

ajvb (approved):

ameihm0912 (approved):

PR #499, opened 2020-11-17
refactor deduplication for aws cross account correlator

kkleemola:
follow up to #496 

#496 resulted in a non updateable graph because of the deduplication transform. I also tried switching it to streaming engine to see as recommended by the error message, but that also failed with no further details, so this simply moves the deduplication to after we've windowed and GBK.

Because we deduplicate after this, it means if the first event we receive is duplicated, it can take a long time to generate the alert as it'll won't happen on the early firing and instead occurs when the window closes and in practice we need to set the session gap to be ~8 minutes. But since the duplicate event issue has been fixed and this is an edge case I don't think the delay is concerning.

### DIFF
--- a/src/main/java/com/mozilla/secops/authprofile/AwsAssumeRoleCorrelator.java
+++ b/src/main/java/com/mozilla/secops/authprofile/AwsAssumeRoleCorrelator.java
@@ -8,12 +8,12 @@ import com.mozilla.secops.parser.EventFilterPayload.StringProperty;
 import com.mozilla.secops.parser.Normalized;
 import com.mozilla.secops.parser.Payload.PayloadType;
 import java.util.ArrayList;
-import org.apache.beam.sdk.transforms.Deduplicate;
+import java.util.HashMap;
+import java.util.Map;
 import org.apache.beam.sdk.transforms.DoFn;
 import org.apache.beam.sdk.transforms.GroupByKey;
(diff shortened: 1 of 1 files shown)


src/main/java/com/mozilla/secops/authprofile/AwsAssumeRoleCorrelator.java
```diff
@@ -96,7 +84,18 @@ public String apply(KV<String, Event> input) {
...
       ArrayList<Event> events = new ArrayList<>();
-      c.element().getValue().forEach(events::add);
+      // deduplicate based on cloudtrail event ID
+      Map<String, Boolean> processedEvents = new HashMap<>();
+      c.element()
+          .getValue()
+          .forEach(
+              e -> {
+                String id = extractCloudTrailEventID(e);
+                if (!processedEvents.containsKey(id)) {
```
ajvb, line 94: Can we log an error or similar if we are seeing duplicates? imo it would be something worth investigating and fixing if it happens again.

ajvb (approved):

src/main/java/com/mozilla/secops/authprofile/AwsAssumeRoleCorrelator.java
kkleemola: Added.

ameihm0912 (approved):

PR #502, opened 2020-12-21
add parser payload metrics

kkleemola:
Adds metrics to the parser functions to give us a summary of what payloads are being ingested so we can improve insight into if for example log messages change and we begin parsing as a raw payload instead of the expected event.


ajvb (approved):
Awesome. Should we set up an alert in Stackdriver on an increase in RAW payload types or is this more for investigation purposes?

kkleemola:
> Awesome. Should we set up an alert in Stackdriver on an increase in RAW payload types or is this more for investigation purposes?

Yep! I'm hoping it's something that we should be able to alert on pretty easily.

PR #525, opened 2021-04-13
filter empty source ips for ua blocklist

kkleemola:
nginx logs often have - for empty. Not filtering null source ip in UA blocklist caused errors trying to key a null source ip after I fixed the previous bug where we weren't normalizing the user agent.


ajvb (approved):

ajvb:
Should we add some test cases for this?

kkleemola:
Yes, oops! Added.

PR #530, opened 2021-05-11
add session based alert suppressor for crit object notifications

kkleemola:
Basically we always update the counter with the new state, suppress the alert and extend the expiry.

Only if it's expired/new alert do we update. Uses a 30min session gap which seems appropriate after reviewing logs.

Alert suppression is more strictly scoped to user + resource + ip.

Note: there's some repeated code because extending AlertSuppressor can't be done cleanly unfortunately.


src/main/java/com/mozilla/secops/alert/AlertSuppressorSession.java
```diff
@@ -0,0 +1,116 @@
...
+import org.apache.beam.sdk.values.KV;
+import org.joda.time.Duration;
+import org.joda.time.Instant;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Alert suppression using session gap based expiry
+ *
+ * <p>Based on AlertSuppressor: creates a suppressor where we extend the expiry each time we stashed
```
ajvb, line 21: The first sentence is a bit hard to understand: `creates a suppressor where we extend the expiry each time we stashed changes suppress an alert for a key`

src/main/java/com/mozilla/secops/alert/AlertSuppressorSession.java
```diff
@@ -0,0 +1,116 @@
...
+      @TimerId("expiryState") Timer counterExpiry) {
+    String key = c.element().getKey();
+    Alert a = c.element().getValue();
+
+    // Prepare a new state value to use if we need to set it later
+    AlertSuppressionState newss = new AlertSuppressionState();
+    newss.key = key;
+    newss.timestamp = a.getTimestamp().toInstant();
+
+    AlertSuppressionState ss = counter.read();
```
ajvb, line 85: I know `ass` isn't a good variable, but neither is `ss` in my opinion. Let's just go with something longer here?

ajvb (approved):

src/main/java/com/mozilla/secops/alert/AlertSuppressorSession.java
kkleemola: I was obeying convention!!
(changed)

PR #535, opened 2021-07-12
add support for ctxsvc - pipeline nginx logs + session limit analysis

kkleemola:
First pass at support needed for ctx svc.

Modifications to nginx parsing to support messages forwarded through pipeline proxy + first pass at a simple session limit heuristic.
I'll likely change the windowing on this as I don't think the aggregating windows are the right option here, but this has been running using prod data for testing and I had wanted to try our existing windowing strategies before adding more.


src/main/java/com/mozilla/secops/httprequest/heuristics/SessionLimitAnalysis.java
```diff
@@ -0,0 +1,284 @@
...
+                    }
+
+                    // Identify if any
+                    // monitored endpoints have exceeded the threshold and use the one with
+                    // the highest request count
+                    // TODO: this will likely need to be changed to allow for
+                    // alerts on multiple paths in the same window
+                    // currently we could miss an alert for an endpoint
+                    // with a low limit, if there's a limit with
+                    // a high monitor
```
ajvb, line 183: Should there be a validation step for this case in the meantime then? Just so this isn't forgotten and then accidentally broken in the future?

src/main/java/com/mozilla/secops/httprequest/heuristics/SessionLimitAnalysis.java
```diff
@@ -0,0 +1,284 @@
...
+
+                      Integer abIdx = indexEndpoint(i.get(1), i.get(0));
+                      if (abIdx != null) {
+                        if (latestRequest == null || t.getMillis() > latestRequest.getMillis()) {
+                          latestRequest = t;
+                        }
+                        // XXX Just pick up the user agent here; with agent variance this could
+                        // result in a different agent being included in the alert than the one
+                        // that was actually associated with the threshold violation, and should
+                        // be fixed.
```
ajvb, line 170: The main problem here being lack of NAT detection?

ajvb (approved):

src/main/java/com/mozilla/secops/httprequest/heuristics/SessionLimitAnalysis.java
```diff
@@ -0,0 +1,284 @@
...
+import org.joda.time.DateTime;
+import org.joda.time.Instant;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Transform for detection of a single source making excessive requests of a specific endpoint
+ * pattern.
+ *
+ * <p>This is based on Endpoint Abuse Analysis but allows for pattern matching and monitor only
```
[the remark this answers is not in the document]
kkleemola: That was sorta my original plan was to create a replacement for endpoint abuse. I'm not going to create a ticket for that just yet cos I think we should probably think a little more broadly about it. Like there's a few heuristics that can probably be combined and in general I think there's work to be done on making them a bit more composable as well as making the mapping to violations more flexible (both in whether something escalates at all and a way to adjust the violation based on perhaps resource or some confidence score). All of this sort of then comes back to config needing a clean up and so I think I'd rather try and create a more defined roadmap and figure out dependencies and then turn those items into tickets rather than just create tickets that likely sit in a backlog.

src/main/java/com/mozilla/secops/httprequest/heuristics/SessionLimitAnalysis.java
kkleemola: I just added a restriction for only accepting one monitored endpoint for right now until its fixed or I have a clearer idea on how multiple violations will be handled.

src/main/java/com/mozilla/secops/httprequest/heuristics/SessionLimitAnalysis.java
kkleemola: Changed the comment a bit here, lack of NAT detection is definitely something I'm a bit worried about as we don't have good variance like when we usually use endpoint abuse. So far in my test job I don't think it's been an issue + we have other guardrails (higher thresholds with the monitor + lower severity violation)

src/main/java/com/mozilla/secops/httprequest/heuristics/SessionLimitAnalysis.java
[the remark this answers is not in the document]
ajvb: Ok totally! Sounds good.

PR #537, opened 2021-07-12
escalate status code rl to iprepd

kkleemola:
Escalates status code violations to iprepd (same violation amount as error code analysis).

### DIFF
--- a/src/main/java/com/mozilla/secops/Violation.java
+++ b/src/main/java/com/mozilla/secops/Violation.java
@@ -69,6 +69,12 @@ public class Violation {
       public String toString() {
         return "violation75";
       }
+    },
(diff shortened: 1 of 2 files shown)


ajvb (approved):
This LGTM, but could you provide me with a bit more context on this? Why didn't it send a violation before? Is this to fix a specific problem?

Don't need to block on providing the context, just want to make sure I'm staying up to date with it so I can provide the best reviews and such.

kkleemola:
It was just because its a new heuristic and I wanted to be able to tune it before escalating. I do want to change this so we have an escalation option like we do with customs, but have been punting on that until I think of how to simplify the config a bit as otherwise I'm just sorta adding to a mess.

PR #547, opened 2021-09-30
PapertrailPull Cloud Function

ajvb:
PapertrailPull is a cloud function that pulls logs from papertrail on a regular schedule and logs them into Stackdriver. From there, they can be pushed to Pub/Sub for consumption by our SIEM

PapertrailPull works by persisting the last or maximum "log event id" into Datastore. Then, when it is next ran, it grabs all log events from the last log id to the current maximum id and persists that.


pwnbus (approved):
Looks good to me!

PR #549, opened 2022-01-07
add alert suppression for pioneer

kkleemola:
Adds alert suppression logic for pioneer so alert doesn't fire continuously if additional data transfer has stopped.


pwnbus (commented):
LGTM!

